PR #18, opened 2017-11-22
Revert "rm creator. Fixes UIIN-30"

zburke:
Reverts folio-org/ui-inventory#16

We do _eventually_ want to remove the `creator` field list, but currently it's still required by the backend. Jumped the gun a bit.


nielserik:
I don't think there's a JIRA issue yet for removing 'creators' from mod-inventory-storage yet. Guess we might bundle it with other changes - like more instance record properties - for a second version of the end-point.

PR #24, opened 2017-11-28
implement Resource Type, Language, and Location filters. Fixes UIIN-32

zburke:


MikeTaylor:
This is sweet work. I feel _a bit_ queasy about the `componentWillUpdate` trick, but overall I think this is a nice validation of the `<FilterGroups>` approach.

PR #45, opened 2017-12-20
Bugfixes itemform

nielserik:


nielserik:
This branch had 'remove-creator' in it, too early for that.

PR #113, opened 2018-03-05
UIIN-102 Cannot enable ui-inventory

hjiebsco:


hjiebsco:
@zburke Is there a guideline for this? Because I also see https://github.com/folio-org/ui-search/blob/master/package.json#L36 and https://github.com/folio-org/ui-eholdings/blob/master/package.json#L67 ?

PR #114, opened 2018-03-06
Change perm display name to be more intuitive

hjiebsco:


hjiebsco:
@MikeTaylor There is already one with that name. Take a look the comments from previous PR to see what's your opinion. https://github.com/folio-org/ui-inventory/pull/113. Seems there is no good established guideline to name permission and use visible attribute.

MikeTaylor:
I'll check after this meeting I'm in. Please hold fire till then. Thanks.

MikeTaylor:
Thanks for your patience.

OK, I see what you're doing -- the `module.inventory.enabled` permission is already in the package file, as it should be, and you're wanting to create a higher-level permission that includes it, together with the necessary back-end permissions.

Here, we run into a much more fundamental problem: one that I'm glad you've raised here, as we approach v1. As far as I know, the people behind the permission model have simply not yet defined any meaningful permissions for any application except Users. I'll bring it to their attention.

The right solution is for Inventory to work the same way as Users: we won't have a single "Able to use the Inventory module" permission, but a selection of high-level permissions analogous to the Users module's "Users: Can view user profile", "Users: Can edit user profile", "Users: Can create new user", etc. And the definition of those sets is what we need from outside.

In the mean time, I do see that you need a do-everything high-level permission, just so you can get on and get your work done. So I suggest that you make its status really clear. Instead of the permission-name `ui-inventory.enabled`, we should go with something like `ui-inventory.all-permissions.TEMPORARY`; and the accompanying user-readable text should say something simily. That way, we won't be able to forget what we've done ... as we gradually did over the last six months to a year!

MikeTaylor:
... and I created https://issues.folio.org/browse/UIIN-103 for discussion of the underlying issue.

nielserik:
Wonder if that's something that the MM SIG are interested in; who can manage meta data and at what levels, like create or edit bib records and holdings records, for instance.

PR #147, opened 2018-04-04
Show availability info on item-details pane

zburke:
Show availability info on the item-details pane, including item status and, if any, links to open requests and the loan details. 

Fixes [UIIN-110](https://issues.folio.org/browse/UIIN-110)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,7 @@
 * Fix Okapi permission names. UIIN-136.
 * Add contributorTypeId and contributorTypeText to instance form. UIIN-92.
 * Ignore yarn-error.log file. Refs STRIPES-517. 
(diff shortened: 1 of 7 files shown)


zburke:
@MikeTaylor, @mkuklis, see the comments on [`getDerivedStateFromProps`](https://github.com/folio-org/ui-inventory/blob/4d8557312a6cd34acc50b509755d5786988e54ad/ViewItem.js#L73-L129). Any idea what's going on there?

mkuklis:
@zburke I had to extract this from your comment. It's worth keeping :)

* dataKey, with a lower case "d",
* that rhymes with "t",
* that stands for "tap dancing" or maybe "tesseract" and also "thelonious"
* that rhymes with "felonious"
* and that stands for "funk!"

PR #169, opened 2018-04-24
handle i18n props correctly. Fixes UIIN-172

zburke:
Fixes [UIIN-172](https://issues.folio.org/browse/UIIN-172)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,7 @@
 * Handle "metaData" field case insensitively for compatibility. Fixes UIIN-159.
 * Only show item-availabilty for open loans. Refs UIIN-110.
 * Add location-units and locations to Okapi interfaces. Fixes UIIN-171.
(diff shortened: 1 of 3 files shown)


MikeTaylor:
What is the difference between the behaviour of the old and the new code?

zburke:
The `props` value handed into `validate` is provided by `redux-form` and does not contain `stripes`. Thus, we must use `<FormattedMessage />` because `props.stripes.intl.formatMessage()` is not available.

MikeTaylor:
Wow, that's a nice subtlety! Thanks for explaining.

zburke:
I take that back, sort of. The `props` value provided to `validate` by `redux-form` contains the same `{...props}` that were passed into `ItemForm`, so the issue this case is that we never passed `stripes` to `ItemForm` from `ViewItem` in the first place.

PR #172, opened 2018-04-24
Favor metadata over metaData. Refs UIIN-173

zburke:
Refs [UIIN-173](https://issues.folio.org/browse/UIIN-173)


zburke:
@funkymalc notes that we may have some out of sync backend module dependencies on folio-snapshot-stable which are causing this merge to fail. Gonna try again, living dangerously without running the regression tests in Jenkins.

PR #181, opened 2018-04-30
Enable opening item view in new tab. UIIN-148

nielserik:


zburke:
This `<Link>` wrapper around the `<Button>` causes the whole app to reload, which is not quite right. I have a feeling the same issue behind https://issues.folio.org/browse/UIIN-122, whatever it is, may also be in play here.

PR #191, opened 2018-05-04
UIIN-156 UIIN-157 UIIN-158: Permissions for Instance Settings

doytch:
Since they're all touching the same things, figured it's easier to knock these together into one PR.

- Added Contributor Types, Resource Types, and Formats pages to the Settings.
- Threw a ControlledVocab on them just for the time being. Future customisation will occur as part of UIIN-150 151 and 152.
- Created the composite permissions for each of those Settings pages.

- Also cleaned up the translations file since there were a bunch of duplicate translations...


doytch:
@zburke I purposely didn't translate the `columnMapping` prop because I thought I remembered some weirdness with sorting around that stuff. Am I remembering correctly? If so, how are we handling these atm? Translate and deal with screwiness for a while or leave it untranslated?

zburke:
You should be okay translating the values in `columnMapping`. The issue you're remembering is https://issues.folio.org/browse/STSMACOM-93 which is related but ultimately a different beast. The problem there is related to the values used by SearchAndSort and MultiColumnList to handle sort.

PR #202, opened 2018-06-06
STCOM-284 Update stripes-components paths

cherewaty:
https://issues.folio.org/browse/STCOM-284

Components previously found in `@folio/stripes-components/lib/structures` are now at `@folio/stripes-components/lib`.


zburke:
Geeez, I'm a total tool. Can't believe I didn't notice this before I merged #207.

PR #277, opened 2018-09-17
Added columnMapping and translated strings for Instances

rasmuswoelk:
We are removing capitalization in CSS for column headers in MultiColumnList which mean that lowercased column headers will be displayed as lowercase instead of being capitalized.

To fix this I added the column translated headers so that the content of these strings can be controlled elsewhere.


rasmuswoelk:
Whoops. Forgot to pull latest changes. I will fix ASAP.

rasmuswoelk:
@zburke Alright! Thanks for letting me know. It's changed now 👍

PR #279, opened 2018-09-17
STCOM-329 Use local RepeatableField

cherewaty:
https://issues.folio.org/browse/STCOM-329

The `stripes-components` `<RepeatableField>` has some breaking changes coming up soon (preview at https://github.com/folio-org/ui-eholdings/pull/545).

`ui-inventory` is the biggest consumer of `<RepeatableField>`. To make the transition to the new `<RepeatableField>` smoother, I've copied the `stripes-components` `<RepeatableField>` into this repo (complete with as much git history as I could move). After the next major `stripes-components` release, the `ui-inventory` uses of `<RepeatableField>` can be migrated one at a time thanks to this PR, instead of having to refactor all of them at once.


MikeTaylor:
> I've copied the stripes-components `<RepeatableField>` into this repo (complete with as much git history as I could move).

How are you doing this? It's something I've often wanted to do, but never figured out a way.

cherewaty:
1. In `stripes-components`: `git-subtree split --prefix=lib/RepeatableField --branch=repeatable-field-subtree`
2. You'll now have a branch that only contains that subtree at the root of the repo. Move the files to what their location will be in their new repo (in this case, `/src/components`)
2. Temporarily add the new repo as a remote: `git remote add ui-inventory https://github.com/folio-org/ui-inventory`
4. Push the `repeatable-field-subtree` branch to the `ui-inventory` remote
5. Switch to a local copy of `ui-inventory`
6. Create `ui-inventory` branch
7. Pull the `repeatable-field-subtree` branch from the `ui-inventory` remote: `git pull origin repeatable-field-subtree --allow-unrelated-histories`

JohnC-80:
@cherewaty some strong git-foo right there!

PR #297, opened 2018-09-26
Reorganize to match stripes-cli blueprint

cherewaty:
Moves around files to match the standard UI module blueprint defined by `stripes-cli`: https://github.com/folio-org/stripes-cli/tree/master/resources/ui-app

Converging to that standard will make future architectural iterations smoother.


src/edit/InstanceForm.js
```diff
@@ -404,8 +404,8 @@ class InstanceForm extends React.Component {
               <Accordion label={<h3>{formatMsg({ id: 'ui-inventory.classifications' })}</h3>} onToggle={this.onToggleSection} open={this.state.sections.instanceSection10} id="instanceSection10">
                 <ClassificationFields classificationTypes={referenceTables.classificationTypes} formatMsg={formatMsg} />
               </Accordion>
-              <Accordion label={<h3>{formatMsg({ id: 'ui-inventory.otherRelatedInstances' })}</h3>} onToggle={this.onToggleSection} open={this.state.sections.instanceSection11} id="instanceSection11">
-              </Accordion>
+              {/* <Accordion label={<h3>{formatMsg({ id: 'ui-inventory.otherRelatedInstances' })}</h3>} onToggle={this.onToggleSection} open={this.state.sections.instanceSection11} id="instanceSection11">
```
wwilsman: Instead of commenting this out for the empty eslint error, couldn't we just make this self-closing?

```
<Accordion {...} />
```

Although I guess I'm not sure why an empty accordion would exist. If just for the heading, could we use the same heading component used in accordions?

Or maybe this is un-intentionally empty and there are future plans to fill this accordion with stuff. In that case, I think commenting this out would be fine

cherewaty: My guess was that it was unintentionally empty, so just commenting it out felt like the least invasive approach.

PR #303, opened 2018-10-01
STCOM-353 Move duplicate to pane header dropdown

cherewaty:
## Purpose
https://issues.folio.org/browse/STCOM-353

Does not yet put "Cancel editing" or "Expand all / Collapse all" options in pane header dropdown.


cherewaty:
Merging, will fix any regression test failures that arise. Glancing over the tests in this repo, we should be ok.

PR #306, opened 2018-10-03
Use STRIPES v1.0.0

JohnC-80:
This PR mainly conforms imports to stripes-v1.0.0 shape and adjusts dependencies.


package.json
```diff
@@ -194,8 +192,11 @@
     "redux-form": "^7.0.3"
   },
   "peerDependencies": {
-    "@folio/stripes-connect": "^3.2.0",
-    "@folio/stripes-core": "^2.11.0",
+    "@folio/stripes-components": "*",
+    "@folio/stripes-form": "*",
+    "@folio/stripes-smart-components": "*",
+    "@folio/stripes-connect": "*",
+    "@folio/stripes-core": "*",
```
cherewaty: Reduce to just `"@folio/stripes": "^1.0.0"`

package.json
```diff
@@ -183,9 +183,7 @@
     "redux": "^4.0.0"
   },
   "dependencies": {
-    "@folio/stripes-components": "^3.1.0",
-    "@folio/stripes-form": "^1.0.0",
-    "@folio/stripes-smart-components": "^1.8.0",
+    "@folio/stripes": "^1.0.0",
```
cherewaty: Should be in `devDependencies` instead of `dependencies`, right?

PR #325, opened 2018-10-26
UIIN-351 BigTest harness

cowboyd:
## Purpose
https://issues.folio.org/browse/UIIN-351

We want to add both unit test coverage to the inventory application, and also add the ability to start the application in mirage mode so that development can be done without having to actually be connected to any kind of server environment.

## Approach

This adds BigTest and Mirage to the inventory app using the emerging standard bigtest setup of models/serializers/factories going into the tests/bigtest directory.

Since the endpoints were being setup against an existing API, we actually tested them explicitly in the tests/bigtest/tests/mirage-test.js directory. This let me gradually standup the endpoints one at a time, verify that I could fetch from both the index and the individual entity and that the chain of models, serializers and endpoints was working at each point.

In addition to the endpoints, the default scenario was populated with about 25 instance records based on actual instances pulled from http://demo.folio.org

## Strategy

There is a lot of work to be done in creating the initial stub that lets you achieve the very basic functionality. After that, we can proceed incrementally to implement features of the harness in order to facillitate more comprehensive tests.

As such, the strategy was to stand up just enough of the endpoints to be able to both view the application and write a basic test that loaded a list of inventory over the network, made sure that it rendered properly, and that clicking on a item in the list loaded the corresponding inventory record.

That's why in this PR only the _read only_ endpoints for most entities have been implemented. At such time as we want to add test coverage for adding inventory or holdings, we can also make the changes to the harness to make that possible.

By the same token, the pagination and CQL filtering have not be implemented yet with the idea being that when we add tests for filtering and pagination, we can also enhance the harness.

## Screenshots

image


wwilsman:
Updated to use the latest stripes-cli mirage and karma changes. The XHR patch won't be necessary once a new release of `stripes-core` is made.

cherewaty:
`stripes-core` release PR https://github.com/folio-org/stripes-core/pull/482

cherewaty:
`stripes-core` `v2.15.5` released.

cherewaty:
Heads up @nielserik - this will change what tests run on Jenkins builds.

PR #335, opened 2018-11-05
Add accordions to Item form page, refs UIIN-275

aditya-matukumalli:


nielserik:
Thank you for this @aditya-matukumalli , I'll start populating the accordions today.

PR #340, opened 2018-11-07
Add instance status setting to inventory settings, refs UIIN-274

aditya-matukumalli:
- Translations
- Update package json with perms


nielserik:
Pulled and tested locally, looks fine :+1:

PR #347, opened 2018-11-09
STCOM-353: Implement Pane header dropdown on Instance and Holdings Details

Yurii-Danylenko:
## Purpose
Add pane header dropdowns to Instance create, edit pages and Holding create, edit pages. Also, replace text of the 'copy' drop-down action items with 'duplicate'.

## Approach
- Create and pass `menuActionItems` prop with items to `Pane` component.
- Add tests to check drop-downs functionality


src/edit/InstanceForm.js
```diff
@@ -177,6 +178,34 @@ class InstanceForm extends React.Component {
...
+    const titleTranslationKey = initialValues.id ? 'ui-inventory.editInstance' : 'ui-inventory.newInstance';
+
+    return (
+      <span data-test-header-title>
+        <FormattedMessage id={titleTranslationKey} />
+      </span>
+    );
+  }
+
+  getPaneHeaderDropdownItems() {
```
MaksymDryha: I'm curious if there is a reason to create a separate method which just returns an array. Why not just save the array to a variable?

PR #353, opened 2018-11-12
UIIN-384 Move data-test- div to fix pane layout

cherewaty:
## Purpose
Fixes https://issues.folio.org/browse/UIIN-384

https://github.com/folio-org/ui-inventory/pull/325 added `<div data-test-instance-details>` to the root of the instance view component to provide an easy point to hang a testing interactor. `<Paneset>`s need their direct children to be `<Pane>`s, so the layout calculations for the instance view were thrown off.

## Approach
Move the `<div>` down one level to be a child of the `<Pane>`.

## Next steps
- `<Pane>` width calculations are relying on JavaScript that's throwing occasional performance warnings in Chrome. That needs some TLC.
- Maybe `<Pane>` should pass on all given props (like a `data-test-*` attribute) to its top-level div? Would've also prevented this issue from being introduced.

### DIFF
--- a/src/ViewInstance.js
+++ b/src/ViewInstance.js
@@ -287,427 +287,426 @@ class ViewInstance extends React.Component {
     );
 
     return (
-      <div data-test-instance-details>
-        <Pane
-          defaultWidth={paneWidth}
-          paneTitle={<span data-test-header-title>{instance.title}</span>}
-          paneSub={instanceSub()}
-          lastMenu={detailMenu}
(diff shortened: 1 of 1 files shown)


cherewaty:
I changed the strategy here a bit. After https://github.com/folio-org/stripes-components/pull/714, we can now apply the `data-test-*` directly to a `<Pane>` and don't have to add an extra `<div>`.

PR #355, opened 2018-11-13
Holding types settings page, refs UIIN-376

aditya-matukumalli:


MikeTaylor:
@aditya-matukumalli @nielserik Was it intentional to add `MD.json` in this commit? It looks to me like it's an auto-generated file.

PR #367, opened 2018-11-21
UIIN-393: Replace actionMenuItems-prop with actionMenu-prop

rasmuswoelk:
We updated the pane action menu implementation for stripes-components to allow for more advanced action menu's. This removes the `actionMenuItems`-prop and replaces it with the `actionMenu`-prop.

In this PR I have replaced all `actionMenuItems`-implementations with the new `actionMenu`-prop.

**Approach:** Added getActionMenu-method that returns an action menu which is then passed to the actionMenu-prop.

I also added a translation string for "Edit item" and removed a duplicate key of "holdingsType" in en.json as well as removing an unused lodash constant that was causing a lint warning.


src/ViewInstance.js
```diff
@@ -156,6 +157,39 @@ class ViewInstance extends React.Component {
     return ref || {};
   }
 
+  getActionMenu({ onToggle }, instance) {
```
cherewaty: It's a little weird that we have to pass that instance around and that it's defined in `render()`, but that's a problem for another day.

src/ViewItem.js
```diff
@@ -216,6 +219,37 @@ class ViewItem extends React.Component {
     this.props.mutator.query.update({ layer: 'copyItem' });
   }
 
+  getActionMenu({ onToggle }, item) {
```
cherewaty: Same oddity here with `item`, but should be another scope of work.

src/edit/InstanceForm.js
```diff
@@ -199,6 +194,19 @@ class InstanceForm extends React.Component {
     );
   }
 
+  getActionMenu() {
+    const { onCancel } = this.props;
+    return (
+      <Fragment>
```
cherewaty: Shouldn't need a `Fragment` here, since there's only one `Button`.

src/edit/holdings/HoldingsForm.js
```diff
@@ -173,6 +168,19 @@ class HoldingsForm extends React.Component {
     });
   }
 
+  getActionMenu() {
+    const { onCancel } = this.props;
+    return (
+      <Fragment>
```
cherewaty: Shouldn't need this `Fragment` either.

src/ViewInstance.js
```diff
@@ -295,19 +328,7 @@ class ViewInstance extends React.Component {
...
-              href: this.craftLayerUrl('edit'),
-              onClick: this.onClickEditInstance,
-            },
-            {
-              id: 'copy-instance',
-              onClick: () => onCopy(instance),
-              label: <FormattedMessage id="ui-inventory.duplicateInstance" />,
-            }
-          ]}
+          actionMenu={(actionMenuProps) => this.getActionMenu(actionMenuProps, instance)}
```
Yurii-Danylenko: Here on every render cycle new function will be generated to be passed to `actionMenu` prop. It can be optimized with moving `(actionMenuProps) => this.getActionMenu(actionMenuProps, instance)` out of the render to a separate method.

src/ViewItem.js
```diff
@@ -291,15 +325,7 @@ class ViewItem extends React.Component {
...
-            actionMenuItems={[{
-              label: formatMsg({ id: 'ui-inventory.editItem' }),
-              href: this.craftLayerUrl('editItem'),
-              onClick: this.onClickEditItem,
-            }, {
-              id: 'clickable-copy-item',
-              onClick: () => this.onCopy(item),
-              label: formatMsg({ id: 'ui-inventory.copyItem' })
-            }]}
+            actionMenu={actionMenuProps => this.getActionMenu(actionMenuProps, item)}
```
Yurii-Danylenko: Same as was mentioned above in `ViewInstance.js`

src/ViewItem.js
```diff
@@ -93,6 +95,7 @@ class ViewItem extends React.Component {
 
     this.craftLayerUrl = craftLayerUrl.bind(this);
     this.cViewMetaData = props.stripes.connect(ViewMetaData);
+    this.getActionMenu = this.getActionMenu.bind(this);
```
Yurii-Danylenko: With modern JS class syntax we can get rid of binding methods. All that is needed is just to write methods in such way: `getActionMenu = () => {}`

src/edit/InstanceForm.js
```diff
@@ -199,6 +194,19 @@ class InstanceForm extends React.Component {
...
+        <Button buttonStyle="dropdownItem" id="cancel-instance-edition" onClick={onCancel}>
```
Yurii-Danylenko: I know there are no conventions about props quantity inside tag element. In my opinion when we have more than one prop it should be placed on separated lines to increase code readability. Does it make sense? What do you think?

cherewaty: I think that's a great convention 👍

src/ViewItem.js
rasmuswoelk: I love arrow functions but I just kinda made it a habit to avoid using them since I read this article (interesting read!):
https://medium.com/@charpeni/arrow-functions-in-class-properties-might-not-be-as-great-as-we-think-3b3551c440b1

The main reasons:
1. They are slower
2. They can hurt testability

But sure – I'll change it into an arrow function since the rest of the methods in this class are declared as arrow functions anyway.

src/edit/InstanceForm.js
rasmuswoelk: I guess my thought was that more actions would be added in the future and having the Fragment would show that it's recommended to use a Fragment for this purpose oppose to a div. But of course, that would probably be a given for any developer extending this functionality in the future so no need to have it there for now.

I'll remove the Fragments! 👍

src/ViewInstance.js
rasmuswoelk: True. The reason why I decided to pass the `instance` constant was that it is generated above like so:

image

Without fully understanding the logic behind this I didn't want to duplicate this logic in the `getActionMenu`-method.

Would you recommend me doing so? Or maybe adding a method for this called `getInstance` that returns the `instance`?

src/ViewItem.js
rasmuswoelk: I will move the logic into the `getActionMenu`-method. 👍

src/ViewItem.js
rasmuswoelk: I'll move it into the `getActionMenu`-method instead.

src/ViewInstance.js
rasmuswoelk: @cherewaty

src/ViewInstance.js
rasmuswoelk: I commented on this below. I would like your input 👍

src/ViewInstance.js
cherewaty: I like @Yurii-Danylenko's suggestion. That should work.

src/ViewItem.js
cherewaty: Yes, arrow functions are the tiniest bit slower at the moment, but I don't expect that to continue to be the case as browser vendors optimize for it. We don't really have to worry about the testability bit, since we're not writing white box component unit tests like described in that post.

src/ViewInstance.js
rasmuswoelk: Thanks – I'll make the change.

src/ViewItem.js
rasmuswoelk: Got it. I changed all binded methods into arrow functions.

rasmuswoelk:
All of the suggested changes should be done now 👍

src/ViewHoldingsRecord.js
```diff
@@ -82,6 +84,7 @@ class ViewHoldingsRecord extends React.Component {
     };
     this.craftLayerUrl = craftLayerUrl.bind(this);
     this.cViewMetaData = props.stripes.connect(ViewMetaData);
+    this.getPaneHeaderActionMenu = this.getPaneHeaderActionMenu.bind(this);
```
cherewaty: One more `bind` we can avoid with a `() => {}`.

Yurii-Danylenko:
I'm wondering if there is any plans to replace actionMenuItems in ItemForm component?

src/ViewHoldingsRecord.js
rasmuswoelk: Whoops. I'll update that ASAP.

rasmuswoelk:
@Yurii-Danylenko Yes. I'll replace that as well in this PR 👍

src/ViewHoldingsRecord.js
rasmuswoelk: Updated.

rasmuswoelk:
I updated the ItemForm as well.

PR #341, opened 2018-11-08
STCOM-395: Add routing for holdings and item creation pages

Yurii-Danylenko:
## Purpose
Add a possibility to go to the holdings and item creation pages by route path

## Approach
- Add `layer=createItem` query param which appears after click on "Add new item" button .
- Bind layer to the item creation page to this query param.
- Move layers components outside of the `ViewInstance` component general markup to be rendered like independent pages

PR #373, opened 2018-11-27
UIIN-400 Bind update method to ViewInstance

cherewaty:
## Purpose
Resolves https://issues.folio.org/browse/UIIN-400

screen shot 2018-11-27 at 2 57 30 pm

## Approach
`this.props` was undefined in `update` because it didn't have the class scope. This regression was probably introduced with https://github.com/folio-org/ui-inventory/pull/341, but I'm not really sure how this ever worked before that.

### DIFF
--- a/src/ViewInstance.js
+++ b/src/ViewInstance.js
@@ -94,7 +94,7 @@ class ViewInstance extends React.Component {
     this.props.mutator.query.update({ layer: 'createHoldingsRecord' });
   }
 
-  update(instance) {
+  update = (instance) => {
     this.props.mutator.selectedInstance.PUT(instance).then(() => {
       this.closeEditInstance();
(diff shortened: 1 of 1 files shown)


Yurii-Danylenko:
> Commit added that replaces the call to `closeEditInstance` with `onCancelHandler` (although I'm confused by that naming, since it's an "onSuccess").

Yeah, it would be better to name it like resetLayerQueryParam, because it doesn't depend on success or failure(is used 2 times as on cancel handler and 1 as on create handler). Sorry for the inconvenience.

cherewaty:
Commit added renaming `onCancelHandler` to `resetLayerQueryParam`. Thanks @Yurii-Danylenko for the suggestion.

PR #410, opened 2018-12-17
STCOM-391 Remove plus sign from translation strings

cherewaty:
## Purpose
https://issues.folio.org/browse/STCOM-391

Related to https://github.com/folio-org/stripes-components/pull/775

## Approach
Instead of having the "+" directly in the translation file, we'll now do:
```
<Icon icon="plus-sign">
  <FormattedMessage id="ui-module.translationId" />
</Icon>
```


cherewaty:
@Yurii-Danylenko I had just ignored that, since the "+" was removed from all English strings, but I added a commit that wraps all the relevant strings in `<Icon icon="plus-sign">`.

PR #425, opened 2019-01-02
Create new request from item record. Fixes UIIN-410

mkuklis:
https://issues.folio.org/browse/UIIN-410

### DIFF
--- a/src/ViewItem.js
+++ b/src/ViewItem.js
@@ -80,7 +80,7 @@ class ViewItem extends React.Component {
     // the top item, sorted by loan-date descending, is a best-effort.
     loans: {
       type: 'okapi',
-      path: 'circulation/loans?query=(itemId==!{itemId}) sortby loanDate/sort.descending&limit=1',
+      path: 'circulation/loans?query=(itemId==!{itemId})&limit=1 sortby loanDate/sort.descending',
       records: 'loans',
     },
     borrowerId: {},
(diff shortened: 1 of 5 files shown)


src/ViewItem.js
```diff
@@ -267,6 +275,20 @@ class ViewItem extends React.Component {
             <FormattedMessage id="ui-inventory.copyItem" />
           </Icon>
         </Button>
+        <Button
+          href={`/requests?itemBarcode=${firstItem.barcode}&layer=create`}
+          onClick={() => {
```
zburke: Do we need the `onClick` handler? Looks like the `href` passes the same attributes to the same URL.

mkuklis: That was my initial thought too but for some reason the `href` was not executing at all. I'm not sure why. Perhaps ActionMenu is catching and preventing the click events somehow.

src/ViewItem.js
```diff
@@ -80,7 +80,7 @@ class ViewItem extends React.Component {
     // the top item, sorted by loan-date descending, is a best-effort.
     loans: {
       type: 'okapi',
-      path: 'circulation/loans?query=(itemId==!{itemId}) sortby loanDate/sort.descending&limit=1',
+      path: 'circulation/loans?query=(itemId==!{itemId})&limit=1 sortby loanDate/sort.descending',
```
aditya-matukumalli: @mkuklis 
image I receive an error popup failing to fetch the loans here.

zburke: Good catch! Query parameters can/should be specified separately from the path, e.g. https://github.com/folio-org/ui-users/blob/aa9e67298b60ef10e68df6e5886d734c87ea2b62/src/Users.js#L83-L91

aditya-matukumalli: @mkuklis It happened when I pulled down the branch and ran it locally.

PR #467, opened 2019-02-04
move AppIcon import to @folio/stripes/core

zburke:
Refs [STCOM-411](https://issues.folio.org/browse/STCOM-411)


Jenkinsfile
```diff
@@ -8,3 +8,4 @@ buildNPM {
   runTestOptions = '--karma.singleRun --karma.browsers ChromeDocker --karma.reporters mocha junit --coverage'
 }
 
+
```
zburke: This may also have failed if #466 hadn't been merged yet. Thanks for nudging Jenkins though.

PR #471, opened 2019-02-08
UIEN-480: Added test running to the Jenkinsfile

Godlevskyi:
## Purpose
This is PR to simply turn on tests running on CI

PR #472, opened 2019-02-08
before calling whenLoaded, define whenLoaded

zburke:
Refs [UIIN-480](https://issues.folio.org/browse/UIIN-480)


zburke:
This resolves the Jenkins failure that is blocking #471 right now.

PR #499, opened 2019-02-14
Replace query resource with react-router history push. Refs STCOM-302

mkuklis:
This PR is an attempt to replace navigation based on query resource with the tools provided by react-router. It should fix issues found in STCOM-302. 

I'm pretty sure the naming in the Location HOC (and perhaps `withLocation` name) could be improved. I wonder if you can come up with a better naming. 

I primary wanted to see if we can get rid of query resource completely and replace it with the set of navigation helpers (perhaps these helpers could be moved up to stripes-core eventually).

### DIFF
--- a/src/ViewHoldingsRecord.js
+++ b/src/ViewHoldingsRecord.js
@@ -26,6 +26,7 @@ import {
 
 import { craftLayerUrl } from './utils';
 import HoldingsForm from './edit/holdings/HoldingsForm';
+import withLocation from './withLocation';
 
 class ViewHoldingsRecord extends React.Component {
   static manifest = Object.freeze({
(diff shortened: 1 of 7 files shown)


src/ViewHoldingsRecord.js
```diff
@@ -846,7 +844,9 @@ ViewHoldingsRecord.propTypes = {
     temporaryLocationQuery: PropTypes.object.isRequired,
   }),
   onCloseViewHoldingsRecord: PropTypes.func.isRequired,
+  updateLocation: PropTypes.func,
+  goTo: PropTypes.func,
```
VictorSoroka1: Should these new props be required? I found the same situation in other files where that props added. Please check other places too.

src/withLocation.js
```diff
@@ -0,0 +1,57 @@
+import omitBy from 'lodash/omitBy';
+import isNil from 'lodash/isNil';
+import React from 'react';
+import PropTypes from 'prop-types';
+import { parse, stringify } from 'query-string';
+import { compose } from 'redux';
+import { withRouter } from 'react-router';
+
+function withLocation(WrappedComponent) {
```
VictorSoroka1: @mkuklis, about your concern about the name. I was thinking about `withHistory`. But eventually, think that `withLocation` might be the more appropriate here as what it really provides is the way to change location.

VictorSoroka1:
@mkuklis, looks great!

src/ViewHoldingsRecord.js
mkuklis: Thanks for catching this @VictorSoroka1!

PR #501, opened 2019-02-15
Updated PaneHeaders in various components

rasmuswoelk:
In this PR I have updated how paneTitles are implemented in various components.

The primary purpose is to improve the design and ensure consistency across all modules.

### View holdings
image
image

### Edit holdings
image
image

### View Item
image
image

### Edit item
image
image

### DIFF
--- a/src/ViewHoldingsRecord.js
+++ b/src/ViewHoldingsRecord.js
@@ -333,26 +333,18 @@ class ViewHoldingsRecord extends React.Component {
           <div data-test-holdings-view-page>
             <Pane
               defaultWidth={this.props.paneWidth}
+              appIcon={<AppIcon app="inventory" iconKey="holdings" />}
               paneTitle={
-                <div
-                  style={{ textAlign: 'center' }}
-                  data-test-header-title
-                >
-                  <AppIcon
-                    app="inventory"
-                    iconKey="holdings"
-                    size="small"
-                  />
-                  <strong>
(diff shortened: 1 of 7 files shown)


src/ViewHoldingsRecord.js
```diff
@@ -333,26 +333,14 @@ class ViewHoldingsRecord extends React.Component {
...
-                    {_.get(holdingsRecord, ['callNumber'], '')}
-                  </strong>
-                  &nbsp;
-                  <div>
-                    <FormattedMessage id="ui-inventory.holdings" />
-                  </div>
-                </div>
+                <span data-test-header-title>
+                  {holdingsRecord.permanentLocationId ? `${holdingsPermanentLocation.name} >` : null}
+                  {' '}
```
zburke: _super_ nitpicky: should we move the space into the interpolated string on the line above? That would cause it to be omitted when there is no location, which seems like the correct behavior.

rasmuswoelk:
@zburke Tried moving the string to the line above but that triggered an linting error 🙄

src/ViewHoldingsRecord.js
zburke: I was thinking `... > ` (note the final trailing space). Not worth fighting about. I _hate_ ESLint's rules about whitespace, TBH, and feel they often do more harm then good.

rasmuswoelk: Ah.. 😅I missed that. Updating that now!

PR #505, opened 2019-02-21
Add option to mark an item as missing, refs UIIN-419

aditya-matukumalli:


src/ViewItem.js
```diff
@@ -240,9 +247,33 @@ class ViewItem extends React.Component {
     this.props.updateLocation({ layer: 'copyItem' });
   }
 
+  handleConfirm = (item, requestRecords) => {
+    const newItem = _.cloneDeep(item);
+    _.set(newItem, ['status', 'name'], 'Missing');
+
+    const newRequestRecord = _.cloneDeep(requestRecords);
```
mkuklis: @aditya-matukumalli can we just clone `requestRecords[0]` here?

aditya-matukumalli: @mkuklis Good catch. Will make the change

package.json
```diff
@@ -300,6 +300,7 @@
     "sinon": "^7.0.0"
   },
   "dependencies": {
+    "@folio/react-intl-safe-html": "^1.0.200010",
```
zburke: This version does not exist outside CI. If there's a bug fix in react-intl-safe-intl we need to depend on, let's get it released and depend on a proper, released version here.

zburke: Actually, NBD, but dealt with in #514 / [UIIN-500](https://issues.folio.org/browse/UIIN-500)

PR #512, opened 2019-02-27
Include Open - In transit status when filtering the requests, refs UI…

aditya-matukumalli:
…IN-358

### DIFF
--- a/src/ViewItem.js
+++ b/src/ViewItem.js
@@ -79,7 +79,7 @@ class ViewItem extends React.Component {
     },
     requests: {
       type: 'okapi',
-      path: 'circulation/requests?query=(itemId==:{itemid}) and status==("Open - Awaiting pickup" or "Open - Not yet filled") sortby requestDate desc',
+      path: 'circulation/requests?query=(itemId==:{itemid}) and status==("Open - Awaiting pickup" or "Open - Not yet filled" or "Open - In transit") sortby requestDate desc',
       records: 'requests',
       PUT: {
         path: 'circulation/requests/%{requestOnItem.id}'


zburke:
IIRC this is the second or third time we've had to update queries in different repositories (inventory, requests, users) to handle new request status options. It makes me wonder if a better way to run the query would be a pattern match on status rather than an exact match, i.e. `=` instead of `==`, something like this:

      path: 'circulation/requests?...status=("Open") sortby requestDate desc',

instead of this:

          path: 'circulation/requests?...status==("Open - Awaiting pickup" or "Open - Not yet filled" or "Open - In transit") sortby requestDate desc',

Thoughts?

aditya-matukumalli:
@zburke That looks like a good option for now, but I am wondering what after we make the change in all the places, if we come across a scenario, for eg: If in the future a new _Open-*_ status is introduced and we have to include all the **Open-*** statuses except this one? I am not sure if there is an easy facility for that using pattern matching?

zburke:
I've asked a few times, "Do you think we've defined all the potential request and loan statuses now?" and each time I hear "Oh, yeah, last one, fer sure." and then a day/week/month later there's a PR like this one to add another status. 

Changing the query from 

    type==(open-foo or open-bar or open-bat)

to

    type=open

would inoculate us from needing to add `open-baz` next month but, as you point out, would prevent us from showing _some_ open statuses but not all. So far, I haven't seen a scenario for that, but it's possible. And if we do see that scenario, we could always roll back to this query style of enumerating all the types. I'm trying to be lazy in the [best sense of the word](http://wiki.c2.com/?LazinessImpatienceHubris) here, i.e. to do the least amount of work possible by doing a little bit more now to save a lot in the future when, inevitably, there is one more type, and another, and another after that :)

aditya-matukumalli:
@zburke Sounds good Zak. As you said I see new issues for example https://issues.folio.org/browse/UIU-682 to do the same. I am inclined as well towards using _
circulation/requests?...status=("Open") sortby requestDate desc',_ instead of the longer version . I can probably start doing this way from UIU-682 and then go through all such instances where the pattern matching hasn't been used under a different JIRA. What do you suggest?

zburke:
I suggest we run it by a PO, @cateboerema since this is her story, and one of the server side folks (@marcjohnson-kint?)

I don't _think_ there should be any issues with this because (a) we're using it like a constant/enumerated type rather than a text wildcard so we should be translating it and (b) since we're already using it to filter results I presume it's already being indexed. But, as above, we should really check with the folks managing the server side to confirm this.

marcjohnson-kint:
@zburke @aditya-matukumalli 

> we're using it like a constant/enumerated type rather than a text wildcard so we should be translating it

Can you provide me with more context about this. What are we discussing translating?

At the moment, most states in the API are referred to by name, rather than another kind of identifier.

> since we're already using it to filter results I presume it's already being indexed

There isn't currently a database index defined for request status. Is that the kind of index being referring to? I don't know if an index would still work effectively using a partial status name.

aditya-matukumalli:
@marc-johnson  The issue we are referring here is to update queries such as 
```
...status==("Open - Awaiting pickup" or "Open - Not yet filled" or "Open - In transit")
```
to
```
...status=(Open) (last I checked this query, it was returning all the open-* statuses from backend)
```
so that we don't have to go around adding a new **open-*** status to the query across different apps every time a new _open-*_ status in introduced in FOLIO. And we can always rollback to the style of enumerating all the status types, if at any point we have to conditionally show some _open_ statuses but not all.

marcjohnson-kint:
@aditya-matukumalli That fits with my understanding of CQL. 

Recent conversations led me to the understanding that the queue is still made up of only open requests. I don't know about the other uses.

zburke:
>> since we're already using it to filter results I presume it's already being indexed
>
> There isn't currently a database index defined for request status. Is that the kind of index being referring to? I don't know if an index would still work effectively using a partial status name.

Yes, that's the type of index I am referring to. If we don't have an index on this column, or if using a partial status name somehow doesn't work for the index, then I start to think we need a new attribute such as `isOpen` for the `status` attribute.

marcjohnson-kint:
@zburke I've been meaning to write up my thoughts about the isOpen property. Has this been raised in a JIRA issue anywhere, or was it only on Slack?

(As you can likely guess, there are trade-offs with storing derived properties)

zburke:
I haven't raised it in a Jira issue.

PR #519, opened 2019-03-05
Replace <IconButton> with <PaneHeaderIconButton>

rasmuswoelk:
We updated stripes-components with an IconButton component that's made specifically for PaneHeaders.

In this PR I replaced the existing IconButton's inside PaneHeaders with the new PaneHeaderIconButton component.


rasmuswoelk:
I just realized that we need to wait until PaneHeaderIconButton is included in a formal release of `stripes-components` before merging this one.

rasmuswoelk:
This PR can't be merged before this project is on stripes ^2.1.0.

PR #530, opened 2019-03-19
UIIN-133: Implement new icons

rasmuswoelk:
## Ticket 
https://issues.folio.org/browse/UIIN-133

## Changes
- Added instance icon for rows in results list and in pane header when viewing instance
- Added item icon for Holding items list when viewing an instance
- Fixed missing space in item pane header
- Minor fix for buttons in Accordion to ensure they align

## Note
It's not possible to add all of the icons specified in the ticket yet because the results list currently only lists instances.

## Screenshots
image

image

image


rasmuswoelk:
Lokos like one or more tests failed. I'll fix those asap.

zburke:
The failure is buried deep in the console log but looks like just a whitespace problem: 
```  ItemViewPage
    visiting the item view page
      ✖ displays the title in the pane header
Chrome 71.0.3578 (Linux 0.0.0) ItemViewPage visiting the item view page displays the title in the pane header FAILED
	AssertionError: expected '337 Item . Available' to equal '337Item . Available'
```

rasmuswoelk:
@zburke Thanks! I'll fix the test now.

src/Items.js
```diff
@@ -48,13 +48,19 @@ class Items extends React.Component {
     if (!items || !items.hasLoaded) return null;
     const itemRecords = items.records;
     const itemsFormatter = {
-      'Item: barcode': x => _.get(x, ['barcode']),
+      'Item: barcode': x => (
+        <span data-test-items-app-icon>
+          <AppIcon app="inventory" iconKey="item" size="small">
+            {_.get(x, ['barcode'])}
```
Yurii-Danylenko: For now, folio supports lodash tree shaking, which allows to import only necessary methods from lodash. Please, import get method as `import { get } from 'lodash'`

Yurii-Danylenko:
@rasmuswoelk, I don't see any test to check that icons exists in results list which is on the first screenshot. Is it possible to cover it?

src/Items.js
zburke: @Yurii-Danylenko, eh, that lodash import hasn't changed in months. I agree we should implement `import { get } from 'lodash'` in new components, but I don't think we need to block PRs for existing components just because they don't make that update.

Yurii-Danylenko: @zburke, no props. Just wanted to apply Boy Scout Rule, to keep the code more or less up to date.

rasmuswoelk: I agree with both of you. It shouldn't block but I appreciate @Yurii-Danylenko's suggestions for making the code better. Since the change is so small I'll go ahead and make it.

Regarding testing of icons in the results list; I'll leave this for another time since it's not apart of the scope of this ticket. I hope that's cool :)

PR #533, opened 2019-03-27
retrieve 1000 rows from lookup tables

zburke:
These hard-coded numbers kill me, but 1,000 is better than 100 I guess
when it comes to retrieving things from lookup tables.

Refs [UIIN-413](https://issues.folio.org/browse/UIIN-413)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change history for ui-inventory
 
+## 1.8.0 (IN PROGRESS)
+
+* Retrieve 1000 items from lookup tables. Refs UIIN-413.
+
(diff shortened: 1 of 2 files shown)


aditya-matukumalli:
@zburke I am assuming this would solve https://issues.folio.org/browse/UIORG-137 as well?

zburke:
This only resolves the problem of locations missing from the filter pane. There are other variations on this problem, e.g. [UIORG-137](https://issues.folio.org/browse/UIORG-137) and [UIIN-411](https://issues.folio.org/browse/UIIN-411) but they all need to be handled separately because the queries are in different places (ui-inventory here, stripes-smart-components for UIIN-411, and ui-organization for UIORG-137).

PR #547, opened 2019-04-16
add missing validation

zburke:
CV fields were missing validation of the required `code` field in many
cases.

CV pages incorrectly passed a `<FormattedMessage>` to the
`labelSingular` field which requires a string; passing a
`<FormattedMessage>` displays `[object Object]` instead.


zburke:
@nielserik, I noticed these validation issues and `[object Object]` message problems with `<FormattedMessage>` when trying to patch a raft of bugs in the `<ControlledVocab>` component. 

Trying to create entries with empty `code` fields would result in a failed request but the error never bubbled up to the UI. This means there's a bug buried deeper in `<ControlledVocab>`, but warning the user about missing required fields as early as possible, _before_ submit, seems polite. 

And `[object Object]` is an obvious bug. 

I didn't find any tickets for this stuff, but it felt wrong to move on without patching it. Can you confirm the required-fields changes, or LMK which PO I should talk to who could?

zburke:
For some reason, SonarCloud [frequently chokes on `ControlledVocab`](/folio-org/ui-users/pull/882).

PR #570, opened 2019-05-06
UIIN-503: Primary Contributor Layout Update

rasmuswoelk:
## Ticket
https://issues.folio.org/browse/UIIN-503

## Changes
- Replaced checkboxes with buttons for contributor fields when editing an instance
- Removed extra asterisk in contributor form
- Added "Primary" column for contributors when viewing an instance

May-06-2019 11-17-42


rasmuswoelk:
@nielserik I accidentally left a `.only` in my tests 🙄It's removed now so it runs all tests again.

src/edit/contributorFields.js
```diff
@@ -74,6 +68,35 @@ const ContributorFields = ({
...
+                  input.onChange(true);
+                };
+
+                return (
+                  <Fragment>
+                    { label && <Label>{label}</Label>}
+                    <Button
+                      data-test-edit-instance-contributors-primary-button
+                      buttonStyle={isPrimary ? 'primary' : 'default'}
+                      onClick={!isPrimary ? handleChange : null}
```
Yurii-Danylenko: Is it possible to move `!isPrimary ? handleChange : null` to a handler method?

src/edit/contributorFields.js
```diff
@@ -74,6 +68,35 @@ const ContributorFields = ({
               name: 'contributorTypeText',
               component: TextField,
             },
+            {
+              name: 'primary',
+              label: intl.formatMessage({ id: 'ui-inventory.primary' }),
+              component: ({ label, meta, input, fields }) => { /* eslint-disable-line react/prop-types */
```
Yurii-Danylenko: What do you think about moving this field to a component?

rasmuswoelk: I think that's a good idea. I'll move it into a new component 👍🏻

src/edit/contributorFields.js
rasmuswoelk: Sure 👍🏻

src/edit/components/PrimaryToggleButton/PrimaryToggleButton.js
```diff
@@ -0,0 +1,56 @@
...
+    onChange(true);
+  };
+
+  return (
+    <Fragment>
+      { label && <Label>{label}</Label>}
+      <Button
+        data-test-primary-toggle-button
+        buttonStyle={isPrimary ? 'primary' : 'default'}
+        onClick={handleChange}
```
Yurii-Danylenko: I believe, `handleClick`, as a handler name, will look a little bit accurate. But it's up to you :)

rasmuswoelk: Makes more sense, yes. I'll change it. Thanks :)

PR #575, opened 2019-05-14
i18n search-field menu

zburke:
Fixes [UIIN-498](https://issues.folio.org/browse/UIIN-498)


zburke:
Replaced by #600, which SonarCloud mysteriously likes better WRT code-coverage. WTF? I don't get it, but I'll take it.

PR #587, opened 2019-05-29
Check for dependencies when deleting holding record. Part of UIIN-550

mkuklis:
https://issues.folio.org/browse/UIIN-550

### DIFF
--- a/src/ViewHoldingsRecord.js
+++ b/src/ViewHoldingsRecord.js
@@ -43,6 +43,12 @@ class ViewHoldingsRecord extends React.Component {
         path: 'holdings-storage/holdings',
       },
     },
+    items: {
+      type: 'okapi',
+      records: 'items',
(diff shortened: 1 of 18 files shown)


src/ViewHoldingsRecord.js
```diff
@@ -352,26 +365,31 @@ class ViewHoldingsRecord extends React.Component {
...
-        <FormattedMessage id="stripes-core.button.back" />
+        <FormattedMessage id="stripes-core.button.cancel" />
       </Button>
     );
 
     return (
       <div>
         <ConfirmationModal
-          data-test-delete-confirmation-modal
+          id="data-test-delete-confirmation-modal"
```
zburke: An `id` of `data-test-...` seems a bit odd. Maybe just `id="delete-confirmation-modal"`?

mkuklis: Good catch @zburke! I replaced it with id because `ConfirmationModal` doesn't take the `data-test-*` but I should have changed the name.

PR #588, opened 2019-05-31
label Save buttons consistently with other apps

zburke:
`Save and close` buttons should say just that and only that, across all
apps, rather than using app-specific language such as `Create new
instance` or `Update holdings record`.

Unrelated, there is no reason to include an `aria-label` on a button
whose only child is the same text, so now we don't.

Refs [UIIN-530](https://issues.folio.org/browse/UIIN-530)


zburke:
Sorry, SonarCloud, you've got it wrong here. I'm not adding new code; I'm removing a bunch of cruft and you've badly misunderstood the situation.

PR #590, opened 2019-06-04
UIIN-559: Update item notes form checkbox to render label vertically

rasmuswoelk:
## Changes
- Updated checkbox in item notes form to render label vertically and updated the column width to take up less space and leave more room for the input fields instead
- Updated RepeatableField component to allow for changing column widths

**Note:** The RepeatableField-component will soon be replaced with the matching component from stripes-components so the change of the RepeatableField-component within UI-Inventory is only temporary. Similar functionality will be possible when using the RepeatableField-component from stripes-components in the future because you will have more control over your row content.

## Before
image

## After
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 1.9.0 (In Progress)
 
 * Check for dependencies when deleting holding record. Part of UIIN-550.
+* Updated checkbox to render label vertically in item notes form (UIIN-559)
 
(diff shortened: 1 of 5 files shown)


rasmuswoelk:
This change requires the latest version of stripes-components. I will verify that this version is available in ui-inventory before merging this PR.

nielserik:
Nice work. I'm adding this new check box label layout to the holdings record form too, while at it.

PR #591, opened 2019-06-04
UIIN-427: Fix inconsistent Record Source for Instance view

taras-tkachenko:


aliaksei-chumakou:
@taras-tkachenko please pay attention that it contradicts with closed ticket https://issues.folio.org/browse/UIIN-353 

> Display "FOLIO" for 'Metadata source' whenever there is no 'sourceRecordFormat' in JSON

taras-tkachenko:
> @taras-tkachenko please pay attention that it contradicts with closed ticket https://issues.folio.org/browse/UIIN-353
> 
> > Display "FOLIO" for 'Metadata source' whenever there is no 'sourceRecordFormat' in JSON

Yes but now Instance records do not contain this field it all so this ticket is deprecated.

PR #593, opened 2019-06-06
UIIN-79: Remove results display by default

YevheniiMaltsev:
## Purpose

Remove results display in case of there are not any search queries or filters in the scope of the [story](https://issues.folio.org/browse/UIIN-79)

## Screenshots

inventory_no_results_by_default

### DIFF
--- a/src/Instances.js
+++ b/src/Instances.js
@@ -6,10 +6,13 @@ import { FormattedMessage } from 'react-intl';
 import {
   stripesShape,
   IntlConsumer,
+  AppIcon,
 } from '@folio/stripes/core'; // eslint-disable-line import/no-unresolved
-import { SearchAndSort } from '@folio/stripes/smart-components';
-import { filters2cql, onChangeFilter as commonChangeFilter } from '@folio/stripes/components';
-import { AppIcon } from '@folio/stripes-core';
+import {
+  SearchAndSort,
+  makeQueryFunction,
+} from '@folio/stripes/smart-components';
+import { onChangeFilter as commonChangeFilter } from '@folio/stripes/components';
 
 import packageInfo from '../package';
 import InstanceForm from './edit/InstanceForm';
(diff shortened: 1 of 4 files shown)


src/Instances.js
```diff
@@ -92,61 +95,38 @@ class Instances extends React.Component {
...
+              const identifierType = resourceData.identifier_types.records.find(type => type.name.toLowerCase() === queryIndex);
+              const identifierTypeId = identifierType ? identifierType.id : 'identifier-type-not-found';
+              queryTemplate = searchableIndex.makeQuery(identifierTypeId);
+            } else {
+              queryTemplate = searchableIndex.makeQuery();
             }
 
-            return cql;
-          },
+            resourceData.query = { ...resourceData.query, ...{ qindex: '' } };
```
SergiyVSergiyenko: What about 

> resourceData.query = { ...resourceData.query, qindex: '' };
?
As for me it looks easier.

src/Instances.js
```diff
@@ -92,61 +95,38 @@ class Instances extends React.Component {
...
+              resourceData,
+              logger
+            ] = args;
+            const queryIndex = resourceData.query.qindex ? resourceData.query.qindex : 'all';
+            const searchableIndex = searchableIndexes.find(idx => idx.value === queryIndex);
+            let queryTemplate = '';
+
+            if (queryIndex === 'isbn' || queryIndex === 'issn') {
+              const identifierType = resourceData.identifier_types.records.find(type => type.name.toLowerCase() === queryIndex);
+              const identifierTypeId = identifierType ? identifierType.id : 'identifier-type-not-found';
```
VictorSoroka1: Please add an empty line after variable declaration.

src/Instances.js
```diff
@@ -92,61 +95,38 @@ class Instances extends React.Component {
...
+              queryTemplate = searchableIndex.makeQuery(identifierTypeId);
```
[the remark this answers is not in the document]
YevheniiMaltsev: Yes, you are right. Thanks!

src/Instances.js
YevheniiMaltsev: Fixed

src/Instances.js
YevheniiMaltsev: Thanks. Fixed

YevheniiMaltsev:
@zburke Thanks for your comment. I agree with you and this is a good option to add the attached solution to solve the issue. And it was agreed to use makeQueryFunction to remove some code duplication and make the code a little bit simpler. It looks like this refactoring worth it because it is easier to support makeQueryFunction instead of partly copied code in every module. 
And I believe that we will be able to make the code base even better in the future.

PR #600, opened 2019-06-12
UIIN-595: Replaced hardcoded translations of search indexes by formatted messages

YevheniiMaltsev:
## Purpose

Replace hardcoded strings of search indexes by formatted messages to support different locales ([issue](https://issues.folio.org/browse/UIIN-595))

## Screenshots

replace_hardcoded_translations_by_formatted_messages


translations/ui-inventory/en.json
```diff
@@ -119,6 +119,9 @@
   "identifier": "Identifier",
   "identifiers": "Identifiers",
   "addIdentifier": "Add identifier",
+  "search.isbn": "- ISBN",
+  "search.issn": "- ISSN",
```
dltj: Please remove the "-" symbol from the strings.  This type of indentation should be done in code...the only thing that should be in the `en.json` string is the label itself.

src/Instances.js
```diff
@@ -59,15 +62,15 @@ const filterConfig = [
...
-  { label: 'Subject', value: 'subject', queryTemplate: 'subjects="%{query.query}"' },
+  { label: 'ui-inventory.search.all', value: 'all', queryTemplate: 'title="%{query.query}" or contributors adj "\\"name\\": \\"%{query.query}\\"" or identifiers adj "\\"value\\": \\"%{query.query}\\""' },
+  { label: 'ui-inventory.barcode', value: 'item.barcode', queryTemplate: 'item.barcode=="%{query.query}"' },
+  { label: 'ui-inventory.instanceId', value: 'id', queryTemplate: 'id="%{query.query}"' },
+  { label: 'ui-inventory.title', value: 'title', queryTemplate: 'title="%{query.query}"' },
+  { label: 'ui-inventory.identifier', value: 'identifier', queryTemplate: 'identifiers adj "\\"value\\": \\"%{query.query}\\""' },
+  { label: 'ui-inventory.search.isbn', value: 'isbn', queryTemplate: 'identifiers == "*\\"value\\": \\"%{query.query}\\", \\"identifierTypeId\\": \\"<%= identifierTypeId %>\\""' },
+  { label: 'ui-inventory.search.issn', value: 'issn', queryTemplate: 'identifiers == "*\\"value\\": \\"%{query.query}\\", \\"identifierTypeId\\": \\"<%= identifierTypeId %>\\""' },
```
dltj: Put the "-" indentation of ISBN and ISSN here instead of in the translation string.

translations/ui-inventory/en.json
YevheniiMaltsev: @dltj thanks for your comments and description in the ticket. The changes were merged to the master branch by @nielserik.
So I will create a new pull request and add you to reviewers.

src/Instances.js
zburke: @dltj, that will get messy. Currently, the entire `label` value is passed as a parameter to `intl.formatMessage()` so adding a prefix like `- ` is not easily supported without changes in a few different places. It's certainly possible, just not trivial.

Would an acceptable approach be for _all_ the `label` values to use `ui-inventory.search...` prefixes? Or is the the feeling that the prefix is formatting, not a translation, and it really doesn't belong in the translation files?

YevheniiMaltsev: @zburke It looks like we can try to use the solution like this #601 to handle this case. What do you think about it?

dltj: @zburke : Oh, nuts.  From a translator's point of view, this issue is similar to the challenge we worked through last year where required fields had the "*" appended to them in the translations strings and we moved that to code.  @YevheniiMaltsev proposed what looks like an interesting approach in #601, but I'm not well versed enough in JavaScript to sign off of the approach.

zburke: #601 LGTM

PR #604, opened 2019-06-13
UIIN-573: Handle display order of Statistical codes different in inst…

VictorSoroka1:
…ance edit dropdown

### DIFF
--- a/src/edit/InstanceForm.js
+++ b/src/edit/InstanceForm.js
@@ -245,13 +245,15 @@ class InstanceForm extends React.Component {
       }),
     ) : [];
 
-    const statisticalCodeOptions = referenceTables.statisticalCodes ? referenceTables.statisticalCodes.map(
-      it => ({
-        label: refLookup(referenceTables.statisticalCodeTypes, it.statisticalCodeTypeId).name + ':    ' + it.code + ' - ' + it.name,
-        value: it.id,
-        selected: it.id === initialValues.statisticalCodeId,
-      }),
-    ) : [];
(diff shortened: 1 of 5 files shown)


src/edit/InstanceForm.js
```diff
@@ -245,13 +245,17 @@ class InstanceForm extends React.Component {
...
-    const statisticalCodeOptions = referenceTables.statisticalCodes ? referenceTables.statisticalCodes.map(
-      it => ({
-        label: refLookup(referenceTables.statisticalCodeTypes, it.statisticalCodeTypeId).name + ':    ' + it.code + ' - ' + it.name,
-        value: it.id,
-        selected: it.id === initialValues.statisticalCodeId,
-      }),
-    ) : [];
+    const statisticalCodeOptions = referenceTables.statisticalCodeTypes
+      .reduce((codes, type) => {
+        return codes.concat(referenceTables.statisticalCodes.filter(code => type.id === code.statisticalCodeTypeId));
```
zburke: My understanding is that we want code-type-name as a secondary sort field after code-name. I don't understand how the `reduce()` step affects sort. Instead of adding a `reduce` step before `map`, wouldn't you just add 
```
.sort((a, b) => return a.label.localeCompare(b.label))
```
after it? If it's just too early in the morning for me to understand this, fine, I'll go drink more coffee. But maybe add a comment explaining how this works then? It isn't obvious what's going on at first glance.

VictorSoroka1: Since `statisticalCodeTypes` are already rearranged `statisticalCodes` to comply with this. Reduce was used for only just that.

PR #611, opened 2019-06-19
UIIN-603: Fix all issues except #1 :: DONE

taras-tkachenko:


doytch:
* I'm not sure I see why `this.instance` is really necessary here. It's going against the spirit of a React app for a single usage in `getMARCRecord` as far as I can see, which could likely be accomplished another way.
* I know you didn't start it, but we really should be cleaning up these wholesale imports of the entire lodash library by just importing what we need, eg, `import { get } from 'lodash')`
* If you're setting `throwErrors: false` to avoid errors when those modules don't exist, instead use the [`fetch`-as-a-function functionality](https://github.com/folio-org/stripes-connect/blob/master/doc/api.md#rest-resources). Eg, `fetch: (props) => !!props.stripes.hasInterface('orders', '6.0'),`
* `console.log` is discouraged. Your usage of it could more appropriately use `console.error`

taras-tkachenko:
Hi @doytch I see your point, thanks.
Donno how to squeeze `instance` to `getMARCRecord` from `render`: `setState()` in `render` is not an option either...

src/ViewInstance.js
```diff
@@ -116,58 +134,82 @@ class ViewInstance extends React.Component {
...
-  }
+  };
 
   closeViewHoldingsRecord = (e) => {
     if (e) e.preventDefault();
     this.props.goTo(`/inventory/view/${this.props.match.params.id}`);
-  }
+  };
+
+  getMARCRecord = (e, instance) => {
```
OleksandrYehorovEPAM: This could look nicer with `async/await`. Not a big deal though.

src/ViewMarc.js
```diff
@@ -10,32 +10,46 @@ import {
...
+    },
   });
 
+  componentDidUpdate() {
+    let { marcRecord } = this.props;
+    const { onClose } = this.props;
+
+    if (!marcRecord) {
+      this.props.mutator.marcRecord.GET()
+        .then(data => { marcRecord = data; })
```
doytch: What is this line supposed to do? I don't see you using the local `marcRecord` variable anywhere else in this scope. It almost looks like you're trying to change the value of `this.props.marcRecord`?

taras-tkachenko: Yeah, my fault, fixed.

src/ViewMarc.js
```diff
@@ -10,32 +10,45 @@ import {
...
+      throwErrors: false,
+    },
   });
 
+  componentDidUpdate() {
+    const { onClose } = this.props;
+
+    if (!this.props.marcRecord) {
+      this.props.mutator.marcRecord.GET()
+        .then(data => { this.props.marcRecord = data; })
```
doytch: This is still an issue. You should never mutate props on a React component. It subverts the entire React philosophy of props-go-down-and-events-go-up. If you want to store something for use locally here, use `state`. If you want the updated `marcRecord` to be available to the parent, use a callback.

taras-tkachenko: I don't need it to be visible to a parent.
It's just a fallback for the case when the user reloads the page with the view record layer open.
In this case MARC Record generated on the parent and given the layer through props is not available any more, so I need to check if it's present and if it's not then try to retrieve it in place.

taras-tkachenko: But I see your point and will try to rework it to state usage.

taras-tkachenko: Done.

src/ViewMarc.js
```diff
@@ -10,32 +10,52 @@ import {
...
+  componentDidMount() {
+    const {
+      marcRecord,
+      mutator: { marcRecord: connection },
+      onClose,
+    } = this.props;
+
+    if (!marcRecord) {
+      connection.GET()
+        .then(data => this.setState(state => ({ ...state, marcRecord: data })))
```
OleksandrYehorovEPAM:
```suggestion
        .then(data => this.setState({ marcRecord: data }))
```
It is not required to spread the state when updating it. React does it under the hood.
https://reactjs.org/docs/state-and-lifecycle.html#state-updates-are-merged

OleksandrYehorovEPAM:
```suggestion
        .then(data => this.setState({ marcRecord: data }))
```
It is not required to spread the state when updating it. React does it under the hood.
And you don't have to pass a function; object is enough.
https://reactjs.org/docs/state-and-lifecycle.html#state-updates-are-merged

PR #617, opened 2019-06-27
UIIN-592: Add field premissions :: DONE

taras-tkachenko:


src/edit/InstanceForm.js
```diff
@@ -203,7 +226,14 @@ class InstanceForm extends React.Component {
         </Icon>
       </Button>
     );
-  }
+  };
+
+  isFieldBlocked = fieldName => {
```
OleksandrYehorovEPAM: Please make it a regular method (not an arrow function) as it does not need to be bound.

src/ViewInstance.js
```diff
@@ -103,6 +109,19 @@ class ViewInstance extends React.Component {
     this.calloutRef = createRef();
   }
 
+  componentDidMount() {
+    this.getMARCRecord();
+  }
+
+  getMARCRecord = () => {
```
OleksandrYehorovEPAM: Please make it a regular method (not an arrow function) as it does not need to be bound.

src/ViewInstance.js
```diff
@@ -203,13 +196,37 @@ class ViewInstance extends React.Component {
     });
   };
 
+  handleViewSource = (e, instance) => {
```
OleksandrYehorovEPAM: Please make it a regular method (not an arrow function) as it does not need to be bound.

src/edit/InstanceForm.js
```diff
@@ -433,9 +469,13 @@ class InstanceForm extends React.Component {
                         {
                           label: <FormattedMessage id="ui-inventory.statisticalCode" />,
                           component: Select,
-                          dataOptions: [{ label: 'Select code', value: '' }, ...statisticalCodeOptions],
+                          dataOptions: [{
+                            label: 'Select code', value: '',
```
OleksandrYehorovEPAM: Please move `value` on a new line.

PR #666, opened 2019-08-06
UIIN-660 link barcode instead of row

zburke:
Link the barcode only, not the entire row.

Fixes [UIIN-660](https://issues.folio.org/browse/UIIN-660)


zburke:
bad rebase pulled in some extra commits; let's try again.

PR #667, opened 2019-08-06
console cleanup

zburke:
Reorganizing so that we could easily pass a string instead of a
`<FormattedMessage>` to `contentLabel` involved wrapping the entire
block returned by `render()` with `<IntlConsumer>`, so we don't need to
wrap individual blocks with it anymore. Wrapping a block with
`<IntlConsumer>` also feels more elegant than with a
`<FormattedMessage>` so those have been cleaned up too.


mkuklis:
That's a very nice pattern @zburke!

PR #668, opened 2019-08-06
UIIN-660 link barcode instead of row

zburke:
Link only the barcode, not the entire row. 

Fixes [UIIN-660](https://issues.folio.org/browse/UIIN-660)


zburke:
Replaces #666 which was, shocker, acting evil.

PR #687, opened 2019-08-16
UIIN-675: ViewSource shows the same record when record changes without ViewRecord closing :: FIXED

taras-tkachenko:
UIIN-675: ViewSource shows the same record when record changes without ViewRecord closing :: FIXED

### DIFF
--- a/src/ViewMarc.js
+++ b/src/ViewMarc.js
@@ -105,10 +105,7 @@ class ViewMarc extends React.Component {
                 <FormattedMessage id="ui-inventory.marcSourceRecord" />
               </h3>
               <div style={{ 'whiteSpace': 'pre', 'fontFamily': 'courier' }}>
-                <table
-                  border="0"
-                  style={{ 'tableLayout': 'fixed' }}
-                >
+                <table style={{ tableLayout: 'fixed', border: 0 }}>
                   <tbody>
                     <tr
                       key="leader"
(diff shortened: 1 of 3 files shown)


src/ViewMarc.js
```diff
@@ -105,10 +106,7 @@ class ViewMarc extends React.Component {
                 <FormattedMessage id="ui-inventory.marcSourceRecord" />
               </h3>
               <div style={{ 'whiteSpace': 'pre', 'fontFamily': 'courier' }}>
-                <table
-                  border="0"
-                  style={{ 'tableLayout': 'fixed' }}
-                >
+                <table style={{ 'tableLayout': 'fixed', 'border': 0 }}>
```
VictorSoroka1:
```suggestion
                <table style={{ tableLayout: 'fixed', border: 0 }}>
```

src/ViewMarc.js
```diff
@@ -29,6 +29,7 @@ class ViewMarc extends React.Component {
     } = this.props;
 
     if (!marcRecord) {
+      this.setState({ marcRecord: null });
```
VictorSoroka1: Not sure this line is needed since `marcRecord` is already set to null [here](https://github.com/folio-org/ui-inventory/pull/687/files#diff-a7ab07b300a26c67bd1523c3d03879f3R22).

src/ViewMarc.js
taras-tkachenko: Done

src/ViewMarc.js
taras-tkachenko: Done

PR #688, opened 2019-08-16
STCOM-554 more lenient selector is more robust

zburke:
In some environments, the CI tests run with a narrow window and that
causes the DOM structure of the `<MultiSelect>` options to change. In
order to accommodate both narrow and wide windows, this PR changes the
selector so that it will succeed in either environment.

Refs [STCOM-554](https://issues.folio.org/browse/STCOM-554)


aditya-matukumalli:
Not sure if this is helpful, but I have countered the narrow screen size issues in tests using _MultiSelect_ by increasing the width of the screen that nightmares options provides

`nightmare.options.width = 1300`

 Usage can be found [here](https://github.com/folio-org/ui-agreements/blob/master/test/ui-testing/tags-crud.js#L14)

PR #691, opened 2019-08-20
console cleanup

zburke:
Copy-number is an array on the backend but exposed as a non-repeatable
field on the front-end. Thus, we need to convert the value when
displaying the form (to provide the expected data-type to the Field,
which would get an array but wants a string/number) just as we don when
saving the form (which would receive a string/number but wants an
array).

### DIFF
--- a/src/ViewItem.js
+++ b/src/ViewItem.js
@@ -400,6 +400,13 @@ class ViewItem extends React.Component {
     return false;
   };
 
+  formValues = item => {
+    const target = Object.assign({}, item);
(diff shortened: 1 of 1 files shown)


doytch:
Obviously I'm not 100% in this code at the moment so it may be impossible, but could we just change the `name` prop to the `Field` that edits this? So that rather than it being `<Field name="copyNumbers" />` it's `<Field name="copyNumbers[0]" />`? Wouldn't that let us skip the conversions?

PR #697, opened 2019-09-06
UIIN-696 allow unassigning of temporary loan type

zburke:
Passing a `placeholder` prop to `<Select>` creates a disabled first
element. Here, we want a placeholder, but we want it enabled because
this field is optional.

Fixes [UIIN-696](https://issues.folio.org/browse/UIIN-696)


doytch:
You may be interested in [STCOM-576](https://issues.folio.org/browse/STCOM-576). Obv it's not useful for fixing this now, but it may be worth linking against/documenting/filing into your memory.

PR #711, opened 2019-09-25
UIIN-725 correctly import Field

zburke:
Forms in ui-inventory were converted from redux-form to
react-final-form, but `<ControlledVocab>` is a stripes-smart-component
form, not a ui-inventory form. Hence, its `<Field>` components must be
imported from `redux-form`.

Fixes [UIIN-725](https://issues.folio.org/browse/UIIN-725)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 1.13.0 (IN PROGRESS)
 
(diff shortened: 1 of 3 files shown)


zburke:
Whoa; the `package.json` update moving `react-router` to peerDependencies should _NOT_ have been part of this PR! [STCOM-581](https://issues.folio.org/browse/STCOM-581).

PR #717, opened 2019-10-02
UIIN-719: Top aligning app icons in the results list

rasmuswoelk:
Added iconAlignment='baseline' prop on AppIcon's in the results list. This ensures that the icon aligns with the first line of the label if there are multiple lines of text.

**Note:** The `iconAlignment`-prop is not yet available. I submitted [a PR](https://github.com/folio-org/stripes-core/pull/728) in stripes-core with an update for the `<AppIcon>`. This PR can be merged before without breaking things.

## Before
This is how iconAlignment="center" aligns the icon (default).

image

## After
This is how iconAlignment="baseline" aligns the icon.

image


rasmuswoelk:
@zburke The`iconAlignment` is a prop on `<AppIcon>` which is located in `stripes-core`. I guess the version bump still applies but I just want to be sure :)

zburke:
@rasmuswoelk, you're right on both accounts. The version bump _is_ still necessary, but it's to pick up the changes from `stripes-core`, not `stripes-components`. Good eye for the details!

PR #726, opened 2019-10-09
Update language filter to support ISO-639 format. Part of UIIN-735

mkuklis:
This PR changes language filter to support ISO-639 format. It also cleans up the previous language list which contained some additional unused data (with over 22k lines of code). The previous list was breaking `MultiSelection` component.

The work here could be potentially moved under https://github.com/folio-org/stripes-smart-components/tree/master/lib/SearchAndSort/components into a reusable `LanguageFilter` but let's keep it here for now since it's already using `MultiSelectionFilter`.


zburke:
Integration tests last night failed due to this PR when the select option's value for `Dutch` changed from `dut` to `nld` which led me to investigate further. The data cleanup bit of this PR removed > 300 languages and also changed the three-letter codes for some. Can you verify that these changes are correct? 

image

The screenshot above shows values for the language-list from folio-testing (post-PR) on the left and folio-clover (pre-PR) on the right. 
* folio-clover contains many additional values: Achinese, Acoli, Adangme, Adygei... 
* folio-clover's "Dutch" code is `dut`; folio-testing's is `nld`, likewise for "Albanian" (`alb` vs `sqi`) and probably others are too. The string `sqi` does not appear anywhere in [the original `languages.js` file](https://github.com/folio-org/ui-inventory/blob/24f74c8fe924077e594873a65a77097bc22f384c/src/data/languages.js). Is this new data from a different source? Do we know which set of values is correct? And can we add some commentary about the source of the data? 
* folio-clover contains invalid `[object Object]` codes for languages, e.g. Aljamía, that have an object instead of a string in the original `code` attribute. That's an obvious bug, though that object's `'-status': 'obsolete',` attribute makes me wonder if we should, in fact, be skipping it.

zburke:
p.s. I moved [the jira ticket](https://issues.folio.org/browse/UIIN-735) from "In review" to "In code review".

PR #727, opened 2019-10-09
UIIN-738: Align the UX display of notes in the Holdings record

Vladyslav-Velytskyi:


src/ViewHoldingsRecord.js
```diff
@@ -269,6 +269,11 @@ class ViewHoldingsRecord extends React.Component {
     return false;
   };
 
+  getHoldingsNotes = (noteTypes, notes) => {
+    return notes
+      .filter((noteType) => noteTypes.find(note => note.holdingsNoteTypeId === noteType.id));
+  };
+
```
mariia-aloshyna: There is an unnecessary return. If the arrow function has only return operator, you can remove it and curly braces.
```suggestion
getHoldingsNotes = (noteTypes, notes) => notes.filter((noteType) => noteTypes.find(note => note.holdingsNoteTypeId === noteType.id));
```

Vladyslav-Velytskyi: Thanks, I'll fix it

PR #729, opened 2019-10-10
UIIN-700: Permissions - Inventory. Item View, edit, create. Item delete(UIIN-708)

annamelnyk:


annamelnyk:
### Description
User has abillity to view, create, edit and delete items up to assigned to him permissions

package.json
```diff
@@ -422,6 +422,35 @@
...
+        "visible": true
+      },
+      {
+        "permissionName": "ui-inventory.item.edit",
+        "displayName": "Inventory: Can view, create and edit item",
+        "subPermissions": [
+          "ui-inventory.item.create",
+          "inventory.items.item.put",
+          "inventory-storage.items.item.get",
+          "inventory-storage.items.collection.get"
```
zburke: It seems like these two might belong under the above permission that includes with "Can _view_ ... item" Are you confident they belong here under "... and edit item" instead?

annamelnyk: Yes, I should replace it to view permission. I'll fix it. Thank you.

PR #738, opened 2019-10-22
UIIN-529: Remove '+' plus signs from buttons

Vladyslav-Velytskyi:


mkuklis:
@Vladyslav-Velytskyi do you know why this has been merged with errors reported from jenkins?

mkuklis:
Ah nevermind it looks like BigTest is failing randomly.

Vladyslav-Velytskyi:
@mkuklis, you are definitely right, locally everything works fine, so I just restarted to complete it, when jenkins checked it I merged, so there is no problem

PR #749, opened 2019-10-28
UIIN-643: Update the display of list of Instance notes

Vladyslav-Velytskyi:


src/utils.js
```diff
@@ -44,3 +44,7 @@ export function parseFiltersToStr(filters) {
 
   return newFilters.join(',');
 }
+
+export function getHoldingsNotes(noteTypes, notes) {
```
mariia-aloshyna: Consider using arrow function

src/ViewInstance.js
```diff
@@ -993,7 +968,38 @@ class ViewInstance extends React.Component {
           onToggle={this.handleAccordionToggle}
           label={<FormattedMessage id="ui-inventory.instanceNotes" />}
         >
-          {layoutNotes(referenceTables.instanceNoteTypes, get(instance, ['notes'], []))}
+          {
+            holdingsNotes.length > 0 && (
```
mariia-aloshyna: Consider using isEmpty() lodash method

PR #750, opened 2019-10-28
UIIN-815: Update the display of list of Subject headings

Vladyslav-Velytskyi:


src/ViewInstance.js
```diff
@@ -518,6 +519,8 @@ class ViewInstance extends React.Component {
       );
     }
 
+    const instanceSubjects = get(instance, 'subjects', '').map(item => { return { value: item } });
```
mariia-aloshyna: There is unnecessary return, use brackets around the object.

Vladyslav-Velytskyi: Thanks

PR #751, opened 2019-10-28
UIIN-706: Display Classification identifier types in alphabetic order

Vladyslav-Velytskyi:
… in detailed view


src/ViewInstance.js
```diff
@@ -518,6 +519,8 @@ class ViewInstance extends React.Component {
       );
     }
 
+    const filteredClassifications = orderBy(get(instance, 'classifications', []), ['classificationNumber', 'classificationTypeId'], ['asc']);
```
mariia-aloshyna: Rename it, please
```suggestion
    const orderedClassifications = orderBy(get(instance, 'classifications', []), ['classificationNumber', 'classificationTypeId'], ['asc']);
```

Vladyslav-Velytskyi: ok

PR #757, opened 2019-10-29
UIIN-642: Update the display of list of Series statements

Vladyslav-Velytskyi:


src/ViewInstance.js
```diff
@@ -518,6 +518,8 @@ class ViewInstance extends React.Component {
       );
     }
 
+    const seriesStatement = instance.series.map(x => ({value: x}));
```
mariia-aloshyna: Add spaces after and before curly brackets, please
```suggestion
    const seriesStatement = instance.series.map(x => ({ value: x }));
```

PR #763, opened 2019-10-30
Another take on UIIN-758 with separate routes. 

mkuklis:
This PR separates each search segment introduces in https://github.com/folio-org/ui-inventory/pull/753 into a separate route. I originally thought it wouldn't be possible to do so with SAS but it is by modifying `packageInfo` passed into SAS:

https://github.com/folio-org/ui-inventory/pull/763/files#diff-fd3611f373f0b1a809dc00c54a815ecbR145-R151

With this PR in place it should be a bit easier to refactor ui-inventory into `SearchAndSortQuery` in the future.

### DIFF
--- a/src/components/FilterNavigation/FilterNavigation.js
+++ b/src/components/FilterNavigation/FilterNavigation.js
@@ -8,7 +8,7 @@ import {
 
 import { segments } from '../../constants';
 
-const FiltersNavigation = ({ segment }) => (
+const FilterNavigation = ({ segment }) => (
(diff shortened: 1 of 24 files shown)


mkuklis:
I feel like this is less clever than what we had before (with a `segment` as a url param) and it moves us slowly to `SearchAndSortQuery` where that little hack can be removed. 

It would be really nice to eventually get to the point where `ui-users` is now but we have to do it incrementally.

PR #775, opened 2019-11-01
Filter item records by material type. Part of UIIN-777.

mkuklis:
Filter item records by material type. Part of UIIN-777.

## Link
https://issues.folio.org/browse/UIIN-777


test/bigtest/network/config.js
```diff
@@ -378,10 +387,8 @@ export default function configure() {
     totalRecords: 0
   });
 
-  this.get('/material-types', {
-    mtypes: [],
-    totalRecords: 0
-  });
+  this.get('/material-types', ({ materialTypes }) => materialTypes.all());
```
VictorSoroka1: I know that BigTest has a shortcut for that use case, but I know that it may now work every.
```suggestion
  this.get('/material-types');
```

mkuklis: Oh I didn't know about that. That's a pretty cool shortcut.

PR #788, opened 2019-11-07
UIIN-414: Fix wrapping cells in Electronic access table

Vladyslav-Velytskyi:


src/ViewInstance.js
```diff
@@ -421,7 +421,7 @@ class ViewInstance extends React.Component {
 
     const electronicAccessRowFormatter = {
       'URL relationship': x => this.refLookup(referenceTables.electronicAccessRelationships, get(x, ['relationshipId'])).name,
-      'URI': x => <a href={get(x, ['uri'])}>{get(x, ['uri'])}</a>,
+      'URI': x => <a href={get(x, ['uri'])} style={{"word-break": "break-word"}}>{get(x, ['uri'])}</a>,
```
mariia-aloshyna: Consider taking out this style to the variable and using it in this line and in 1142

PR #792, opened 2019-11-11
UIIN-828: Item record. Display empty elements with the dashes

mariia-aloshyna:
**Description**
Item record. Consistent display with a dash in all cases where the element is empty. When an accordion has only empty data elements - marked with a dash, then the accordion by default is to be collapsed.

**Approach**
Extract object variables for accordions. Make a function for checking if an accordion has only empty data elements.

**Issue**
https://issues.folio.org/browse/UIIN-828


src/ViewItem.js
```diff
@@ -844,286 +978,364 @@ class ViewItem extends React.Component {
...
                   </Row>
                   <br />
                   <Row>
-                    <Col smOffset={0} sm={4}>
-                      {get(item, ['effectiveLocation', 'name'], '')}
+                    <Col
+                      smOffset={0}
+                      sm={4}
+                    >
+                      {checkIfElementIsEmpty(itemLocation.effectiveLocation)}
```
Vladyslav-Velytskyi: This is just title, not field with value

mariia-aloshyna: Are you sure? I see that the title is placed on line 1223.

Vladyslav-Velytskyi: Ops, sorry, I looked to another line, now it's LGFM.

PR #810, opened 2019-11-19
UIIN-828: Display a dash if the element is empty

mariia-aloshyna:
Display all fields on the View item screen.
Display a dash if the element is empty.
Display MultiColumnList with dashes if content is not provided.


src/ViewItem.js
```diff
@@ -728,7 +725,44 @@ class ViewItem extends React.Component {
...
+      acc08: areAllFieldsEmpty(Object.values(electronicAccess)),
+    };
+
+    const statisticalCodeContent = (!isEmpty(administrativeData.statisticalCodeIds)
+      ? administrativeData.statisticalCodeIds.map(id => ({ codeId: id }))
+      : [{ codeId: '-' }]);
+
+    const electronicAccessContent = (!isEmpty(electronicAccess.electronicAccess)
+      ? electronicAccess.electronicAccess
+      : [{
```
Vladyslav-Velytskyi: You can combine, string 718, method get can have default value, add as default array this one

mariia-aloshyna: If I add returned object from electronicAccessContent as a default value to electronicAccess field, it will not pass the "checkIfElementIsEmpty()" check (which check if array is empty) and then "Electronic access" accordion will be opened though it contains an empty table.

Vladyslav-Velytskyi: You can omit this comment, actually, if _.get receive form backend an empty array, so defaultValue would be missed

PR #817, opened 2019-11-20
Add search options to holdings segment. Part of UIIN-848.

mkuklis:
https://issues.folio.org/browse/UIIN-848


mkuklis:
I'm not sure why SonarCloud reports 0% here. The tests are there and they should be at lest 80% for this PR.

zburke:
0% because tests are turned off in `Jenkinsfile` so no test coverage data is generated as part of the PR.

mkuklis:
Ah that makes sense. Thanks @zburke! I keep forgetting that they are completely turned off :(

PR #830, opened 2019-11-26
eight is string but 8 is an int

zburke:
Provide correct proptype to `<Col>`.


zburke:
@mkuklis took care of this in #835. He's good like that.

PR #832, opened 2019-11-26
UIIN-884: Display complete call number in holdings headers

Baroquem:
https://issues.folio.org/browse/UIIN-884

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -160,4 +160,11 @@ export const getDate = dateValue => {
   ) : '-';
 };
 
-export const callNumberLabel = hr => `${hr.callNumberPrefix || ''} ${hr.callNumber || ''} ${hr.callNumberSuffix || ''}`;
+export const callNumberLabel = holdingsRecord => {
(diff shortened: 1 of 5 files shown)


src/utils.js
```diff
@@ -159,3 +159,5 @@ export const getDate = dateValue => {
     />
   ) : '-';
 };
+
+export const callNumberLabel = hr => `${hr.callNumberPrefix || ''} ${hr.callNumber || ''} ${hr.callNumberSuffix || ''}`;
```
mkuklis: @Baroquem I think this could create double spaces in some cases right? For example if the `callNumber` is empty we would get two spaces between `callNumberPrefix` and `callNumberSuffix`.

I wonder if a different pattern for this could be something similar to what @zburke did here:

https://github.com/folio-org/stripes-util/pull/20/files#diff-7e21da0a0fe2f8e96c861074618f38a5

Baroquem: Yes, good idea! That's much better.

PR #848, opened 2019-12-05
UIIN-883 - FE: In transit report: check in fields

maximdidenkoepam:
# Purpose
Change path to checkin fields in inTransit.csv report

# Link
https://issues.folio.org/browse/UIIN-883

# Dependency
https://issues.folio.org/browse/CIRC-575


maximdidenkoepam:
Do not merge until CIRC-575 will be merged

PR #857, opened 2019-12-13
UIIN-749: Show most recent check in on item record

SergiyVSergiyenko:
## Purpose
Show most recent check in on item record. [Story](https://issues.folio.org/browse/UIIN-749).

**Screenshot**

Circulation history

### DIFF
--- a/src/ViewItem.js
+++ b/src/ViewItem.js
@@ -92,6 +92,30 @@ class ViewItem extends React.Component {
       },
       records: 'loantypes',
     },
+    servicePoints: {
+      type: 'okapi',
+      path: 'service-points',
+      records: 'servicepoints',
+      params: (_q, _p, _r, _l, props) => {
+        const servicePointId = get(props.resources, 'items.records[0].lastCheckIn.servicePointId', '');
+
+        if (!servicePointId) return {};
+
+        return { query: `id==${servicePointId}` };
+      },
+    },
+    staffMembers: {
+      type: 'okapi',
+      path: 'users',
+      records: 'users',
+      params: (_q, _p, _r, _l, props) => {
+        const staffMemberId = get(props.resources, 'items.records[0].lastCheckIn.staffMemberId', '');
+
+        if (!staffMemberId) return {};
+
+        return { query: `id==${staffMemberId}` };
+      },
+    },
     callNumberTypes: {
       type: 'okapi',
       path: 'call-number-types',
@@ -143,6 +167,7 @@ class ViewItem extends React.Component {
         acc06: true,
         acc07: true,
         acc08: true,
+        acc09: true,
       },
       loan: null,
       borrower: null,
@@ -461,6 +486,8 @@ class ViewItem extends React.Component {
         loanTypes,
         requests,
         callNumberTypes,
+        staffMembers,
+        servicePoints,
       },
       referenceTables,
       okapi,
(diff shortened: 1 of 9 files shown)


src/ViewItem.js
```diff
@@ -179,6 +189,29 @@ class ViewItem extends React.Component {
...
   }
 
+  async componentDidUpdate(prevProps) {
+    if (prevProps.resources.items && this.props.resources.items.records[0] !== prevProps.resources.items.records[0]) {
+      const { lastCheckIn } = this.props.resources.items.records[0];
+
+      if (lastCheckIn) {
+        const { staffMemberId, servicePointId } = lastCheckIn;
+
+        const [staffMembers, servicePoints] = await Promise.all(
```
YevheniiMaltsev:
```
        const [staffMembers, servicePoints] = await Promise.all(
          [
            this.fetchEntityById(staffMemberId, 'borrowers'),
            this.fetchEntityById(servicePointId, 'servicePoints'),
          ]
        );
```

It seems, it will be better to format:

```
        const [staffMembers, servicePoints] = await Promise.all([
          this.fetchEntityById(staffMemberId, 'borrowers'),
          this.fetchEntityById(servicePointId, 'servicePoints'),
        ]);
```

test/bigtest/interactors/item-view-page.js
```diff
@@ -26,6 +27,10 @@ import MultiColumnListInteractor from '@folio/stripes-components/lib/MultiColumn
   clickDelete = clickable('[data-test-inventory-delete-item-action]');
 }
 
+@interactor class CirculationHistory {
+  kvValues = collection('[data-test-kv-value]');
```
YevheniiMaltsev: It seems, kvValues is not a clear name. I know it's used in a lot of places, but maybe we can try give more valuable name.

@VictorSoroka1  what do you think about it?

src/ViewItem.js
```diff
@@ -179,6 +189,27 @@ class ViewItem extends React.Component {
     }
   }
 
+  async componentDidUpdate(prevProps) {
+    if (prevProps.resources.items && this.props.resources.items.records[0] !== prevProps.resources.items.records[0]) {
```
VictorSoroka1: Please move this logic to a function with the meaningless name and call it inside `componentDidUpdate`.

src/ViewItem.js
```diff
@@ -179,6 +189,27 @@ class ViewItem extends React.Component {
...
+      const { lastCheckIn } = this.props.resources.items.records[0];
+
+      if (lastCheckIn) {
+        const { staffMemberId, servicePointId } = lastCheckIn;
+
+        const [staffMembers, servicePoints] = await Promise.all([
```
VictorSoroka1: Please try to take advantage of params prop in the manifest in order to avoid the call in the `componentDidUpdate`. See the [example](https://github.com/folio-org/ui-agreements/blob/master/src/routes/AgreementViewRoute.js#L56) for reference.
 @doytch, could you please suggest whether it makes sense?

doytch: Yeah I think that makes sense @VictorSoroka1. I've definitely been much happier with the functional `params` approach over this `componentDidUpdate` approach in our apps. Would let you avoid a lot of the logic here in this case and make things easier to debug and change in the future.

PR #862, opened 2019-12-17
UIIN-749: Show most recent check in on item record

SergiyVSergiyenko:
## Purpose
Show most recent check in on item record. [Story](https://issues.folio.org/browse/UIIN-749).

**Screenshot**

Circulation history

### DIFF
--- a/src/ViewItem.js
+++ b/src/ViewItem.js
@@ -74,6 +74,7 @@ class ViewItem extends React.Component {
       type: 'okapi',
       path: 'inventory/items/:{itemid}',
       POST: { path: 'inventory/items' },
+      resourceShouldRefresh: true,
     },
     holdingsRecords: {
       type: 'okapi',
@@ -99,10 +100,11 @@ class ViewItem extends React.Component {
       params: (_q, _p, _r, _l, props) => {
         const servicePointId = get(props.resources, 'items.records[0].lastCheckIn.servicePointId', '');
 
-        if (!servicePointId) return {};
+        const query = servicePointId && `id==${servicePointId}`;
 
-        return { query: `id==${servicePointId}` };
+        return query ? { query } : null;
       },
+      resourceShouldRefresh: true,
     },
     staffMembers: {
       type: 'okapi',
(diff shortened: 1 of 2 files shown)


YevheniiMaltsev:
This is an additional fix to #857

src/ViewItem.js
```diff
@@ -99,10 +100,11 @@ class ViewItem extends React.Component {
       params: (_q, _p, _r, _l, props) => {
         const servicePointId = get(props.resources, 'items.records[0].lastCheckIn.servicePointId', '');
 
-        if (!servicePointId) return {};
+        const query = servicePointId && `id==${servicePointId}`;
 
-        return { query: `id==${servicePointId}` };
+        return query ? { query } : null;
```
doytch: So, this is totally fine in normal use and it's actually the way I wrote my code at first. 

I was later writing some BigTests and ran into issues in my Mirage code where it didn't nicely handle a `null` search string. This is why I've been transitioning my code to a `return query ? { query } : {}` pattern instead.

There may be a better way of writing Mirage handlers that obviates the need for returning an empty object here, in which case I'd recommend using your current, `null`-returning pattern instead. Or if you're not planning on using BigTest w/ Mirage handlers for this endpoint, then feel free to ignore this as well.

src/ViewItem.js
```diff
@@ -74,6 +74,7 @@ class ViewItem extends React.Component {
       type: 'okapi',
       path: 'inventory/items/:{itemid}',
       POST: { path: 'inventory/items' },
+      resourceShouldRefresh: true,
```
doytch: Why do you need this here? Our experience so far is that `resourceShouldRefresh` is generally a bit of a hack or a workaround for some other behaviour. As a result, it's always preferable to not rely on it. Barring that, I've found it immensely helpful when people document via a comment _why_ they're setting this to true. It tends to make future changes much simpler and accidental bugs much less likely.

Same comment about Line 107.

src/ViewItem.js
SergiyVSergiyenko: Thanks for your advice about using BigTest with Mirage handlers. I will take this case in mind.
The thing is using an empty-object-returning pattern here, when the entity ID is missing, leads to the receiving of all entities. While using a null-returning pattern does not send a request at all, which meets the requirements of this story.

src/ViewItem.js
zburke: In my experience, `resourceShouldRefresh` gets inserted whenever a PO's test pattern is "create a new entry for lookup-item-X in foo-settings, go to foo, edit a Foo, add comment to Jira that 'lookup-item-X was not present on Foo-edit page until I refreshed the browser'".

SergiyVSergiyenko: Yes, I agree with Zak. Without applying the resourceShouldRefresh field, resource properties are not updated. Thus, when moving to the next item without reloading the page, the form displays the data of the previous item.

PR #863, opened 2019-12-18
UIIN-904: Set fixed width for input fields

mariia-aloshyna:


src/settings/HRIDHendling/HRIDHandling.css
```diff
@@ -16,6 +16,15 @@
    padding-left: 1rem;
 }
 
+.inputLabel {
+  width: 9rem;
+  padding: 0 .5rem;
```
Vladyslav-Velytskyi:
```suggestion
  padding: 0 0.5rem;
```
it could be more readable

PR #869, opened 2019-12-20
UIIN-913: Fix display of information when the item's status is 'Missing'

SergiyVSergiyenko:
## Purpose
Fix display of information in the "Loan and availability" section when the item's status is "Missing". [Story](https://issues.folio.org/browse/UIIN-913).

### Screenshot
Item's missing

### DIFF
--- a/src/ViewItem.js
+++ b/src/ViewItem.js
@@ -4,7 +4,6 @@ import {
   cloneDeep,
   set,
   omit,
-  includes,
   map,
   isEmpty,
   values,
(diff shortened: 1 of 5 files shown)


src/ViewItem.js
```diff
@@ -134,18 +133,13 @@ class ViewItem extends React.Component {
...
       params: {
-        query: 'itemId==!{itemId} sortby loanDate/sort.descending',
-        limit: '1',
+        query: 'status.name=="Open" and itemId==!{itemId}',
       },
       records: 'loans',
-      accumulate: 'true',
-      fetch: false,
     },
     borrowers: {
```
VictorSoroka1: As I understand `borrowers` entry can be deleted at all.

PR #882, opened 2020-01-10
UIIN-902: Add missing translations

mariia-aloshyna:


zburke:
This was added already in #879, but I'm not sure why as I don't see that key in any source file. There's not really an explanation [on UIIN-902](https://issues.folio.org/browse/UIIN-902) of what needs to happen here. Topping it all off, UIIN-902 is already closed. 

So ... can this be closed to?

mariia-aloshyna:
Hi, @zburke Yes, this can be closed, thank you!
This translation is used for rendering FOLIO records' field lists for the Data import module (UIDATIMP-330)

zburke:
> This translation is used for rendering FOLIO records' field lists for the Data import module (UIDATIMP-330)

If that key is for ui-data-import then why is it in ui-inventory?

mariia-aloshyna:
> > This translation is used for rendering FOLIO records' field lists for the Data import module (UIDATIMP-330)
> 
> If that key is for ui-data-import then why is it in ui-inventory?

Almost all the translations we need are already present in other modules, we import these translations and use them in ui-data-import (list of records' fields is attached to the UIDATIMP-330)

PR #896, opened 2020-01-24
UIDEXP-12: Disable saving instances UIIDs option if there are not items in search result

YevheniiMaltsev:
## Purposes

Disable saving instances UIIDs option if there are not items in search result in the scope of the [ticket](https://issues.folio.org/browse/UIDEXP-12).

## Screenshots
disabled_save_UIIDs_button

enabled_save_UIIDs_button

### DIFF
--- a/test/bigtest/tests/export-to-csv-test.js
+++ b/test/bigtest/tests/export-to-csv-test.js
@@ -9,9 +9,10 @@ import sinon from 'sinon';
 
 import setupApplication from '../helpers/setup-application';
 import InventoryInteractor from '../interactors/inventory';
+import InstancesRouteInteractor from '../interactors/routes/instances-route';
 
 describe('Instances', () => {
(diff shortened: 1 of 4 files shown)


test/bigtest/tests/export-to-csv-test.js
```diff
@@ -21,6 +22,21 @@ describe('Instances', () => {
   let xhr;
   let requests = [];
 
+  describe.only('searching by instance HRID to fill items list', function () {
```
mkuklis: @YevheniiMaltsev please remove only here.

VictorSoroka1: Good catch @mkuklis.

YevheniiMaltsev: @mkuklis thanks!

PR #908, opened 2020-02-07
UIIN-954: Add "Volume" and "Year, caption" fields to Instance detaile…

SergiyVSergiyenko:
## Purpose
Add "Volume" and "Year, caption" fields to the Instance record in a detailed view. Story [UIIN-954](https://issues.folio.org/browse/UIIN-954).

### Screenshot
with_Volume_ _Year_wrapped


src/Items.js
```diff
@@ -139,6 +141,8 @@ class Items extends React.Component {
       'materialType': x => get(x, ['materialType', 'name']) || noValue,
       'enumeration': x => x.enumeration || noValue,
       'chronology': x => x.chronology || noValue,
+      'volume': x => x.volume || noValue,
+      'yearCaption': x => x.yearCaption?.join(', ') || noValue,
```
annamelnyk: Looks like in this string is misspelling, `?` should be deleted. @SergiyVSergiyenko could you fix it please?

SergiyVSergiyenko: @annamelnyk, thanks for reviewing. Please take a look at [Optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining).

annamelnyk: @SergiyVSergiyenko sorry for confusing, looks like the problem was with babel version and build showed error in this place. And thank you for link, I did not know about `?.` operator

PR #918, opened 2020-02-13
UIIN-955: Init work on performance improvements

mkuklis:
This PR includes some initial work related to performance improvements:

* Cleanup unnecessary manifests which caused same requests executing multiple times. 
* Refactored routing to a single `InstanceRoute` and  moved filter `segment` into a regular query param. This allowed for removing some code duplication around items route, holdings route and instances route and it will simplify introducing new routes in the future.
* Added `memo` to `InstancesView` to help with unnecessary re-renders. 
* Extract `ItemView` into a separate route

https://issues.folio.org/browse/UIIN-955


src/views/InstancesView.js
```diff
@@ -1,6 +1,9 @@
-import React from 'react';
+import React, { memo } from 'react';
```
Baroquem: Michal, I've never used memo, but my understanding is that it can only be applied to a functional component, which InstancesView is not. Am I wrong?

mkuklis: @Baroquem it should also work with classes: https://github.com/facebook/react/issues/13937

Baroquem: Ok then! I like how they basically say yes, it works, but we don't want to document it in case people start using it.

SergiyVSergiyenko: And what if we simply convert InstancesView into a functional component for consistency?

mkuklis: I like that @SergiyVSergiyenko! I will try to refactor a bit tomorrow.

SergiyVSergiyenko:
Great job! Creatively! Functional programming in action!

src/components/ItemFilters/itemFilterRenderer.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import { get } from 'lodash';
+
+import ItemFilters from './ItemFilters';
+import { getCurrentFilters } from '../../utils';
+import { itemStatuses } from '../../constants';
+
+const itemFilterRenderer = data => onChange => {
```
zburke: `itemFilterRenderer` is a function that takes a single argument (`data`) and returns a function that takes a single argument (`onChange`), right? Cool cool. Maybe add a comment to that effect so it's more obvious to folks like me how to use it without having to stare at it for quite so long? :)

src/components/index.js
```diff
@@ -1,6 +1,6 @@
 export { default as FilterNavigation } from './FilterNavigation';
 export { default as RepeatableField } from './RepeatableField';
-export { default as InstanceFilters } from './InstanceFilters';
-export { default as HoldingFilters } from './HoldingFilters';
-export { default as ItemFilters } from './ItemFilters';
+export { default as InstanceFilters, instanceFilterRenderer } from './InstanceFilters';
+export { default as HoldingFilters, holdingFilterRenderer } from './HoldingFilters';
```
zburke: `HoldingFilters` should maybe/probably be renamed `HoldingsRecordFilters` for consistency.

src/routes/InstancesRoute.js
```diff
@@ -72,20 +79,26 @@ class InstancesRoute extends React.Component {
...
     }
 
+    const data = getData();
+
     return (
       <InstancesView
         parentResources={resources}
         parentMutator={mutator}
-        data={getData()}
+        data={data}
```
zburke: What's the benefit in separating `data` into a variable?

src/components/ItemFilters/itemFilterRenderer.js
mkuklis: Yeah it's hard to read. Still looking for ways to simplify it.

src/routes/InstancesRoute.js
mkuklis: This was a left over from my debugging data :) Will clean it up. Thanks for catching it!

PR #930, opened 2020-02-26
UIIN-990: Filter instance by call number via items segment

mkuklis:
https://issues.folio.org/browse/UIIN-990

### DIFF
--- a/test/bigtest/network/config.js
+++ b/test/bigtest/network/config.js
@@ -122,25 +122,26 @@ export default function configure() {
         right,
       } = cqlParser.tree;
 
-      if (left && right) {
-        if (left.field === 'title' && right.field === 'contributors') {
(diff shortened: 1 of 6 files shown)


test/bigtest/network/config.js
```diff
@@ -122,9 +122,18 @@ export default function configure() {
         right,
       } = cqlParser.tree;
 
-      if (left && right && left.field === 'title' && right.field === 'contributors') {
-        return instances.all().filter(inst => inst.title === left.term &&
-          inst.contributors[0].name === right.term);
+      if (left && right) {
+        if (left.field === 'title' && right.field === 'contributors') {
```
SergiyVSergiyenko: @mkuklis, I wonder if we could use an optional chaining here instead of a nested 'if'?
```suggestion
        if (left?.field === 'title' && right?.field === 'contributors') {
```

PR #935, opened 2020-03-06
UIDEXP-32: Add Save icon to the Save Instances UUID

Yuliia-Dovhal:
## Purpose:
Add Save icon to the Save Instances UUIDs option in Action menu

[UIDEXP-32](https://issues.folio.org/browse/UIDEXP-32)

## Screenshot:
Screenshot 2020-03-10 at 11 17 40


src/components/InstancesList/InstancesList.js
```diff
@@ -15,7 +15,7 @@ import {
 
 import { AppIcon } from '@folio/stripes/core';
 import { SearchAndSort } from '@folio/stripes/smart-components';
-import { Button } from '@folio/stripes/components';
+import { Button, Icon } from '@folio/stripes/components';
```
YevheniiMaltsev:
```suggestion
import {
  Button,
  Icon,
} from '@folio/stripes/components';
```

src/components/InstancesList/InstancesList.js
```diff
@@ -193,6 +195,11 @@ class InstancesView extends React.Component {
             this.generateInstancesIdReport();
           }}
         >
+          <Icon
```
YevheniiMaltsev: I think we can try to add a test to check if the correct (save) icon is in place. @VictorSoroka1  @Yuliia-Dovhal  what do you think about it?

src/components/InstancesList/InstancesList.js
```diff
@@ -15,7 +15,7 @@ import {
 
 import { AppIcon } from '@folio/stripes/core';
```
YevheniiMaltsev: Please, add a screenshot to the pull request description

src/components/InstancesList/InstancesList.js
Yuliia-Dovhal: The existed tests are broken and I will spend too much time for fixing all that tests before even adding my ones. Also it seems like fixing existing tests is not only my responsibility.
Maybe @VictorSoroka1 can disagree?

VictorSoroka1: Agree.

PR #941, opened 2020-03-10
UIIN-950: Add support for unconnected preceding titles

mkuklis:
## Purpose
Add support for unconnected preceding titles.

## Link
https://issues.folio.org/browse/UIIN-950

## Screenshot
titles


SergiyVSergiyenko:
A cool name is marshalTitles. :)  A synonym for “connected”?

mkuklis:
Hahah yes it sounds funny. I was looking for something else than `parse` and `marshalling` sounded like a good fit: https://en.wikipedia.org/wiki/Marshalling_(computer_science)

PR #951, opened 2020-03-16
STCOM-654: Add parse to discoverySuppress and staffSuppress filters.

mkuklis:
This PR uses the new `parse` function introduced in https://issues.folio.org/browse/STCOM-654

to convert `discoverySuppress` and `staffSuppress` filters to CQL manually.

This should address issues described in:

https://issues.folio.org/browse/UIIN-968
https://issues.folio.org/browse/UIIN-969
https://issues.folio.org/browse/UIIN-970


src/filterConfig.js
```diff
@@ -4,6 +4,21 @@ import {
   itemFilterRenderer,
 } from './components';
 
+// Function which takes a filter name and returns
+// another function which can be used in filter config
+// to parse a given filter into a CQL manually.
+const parseFilter = name => values => {
```
zburke: Ooooh! Thank you for the comments!

How about `generateOptionalBoolean` instead of `parseFilter`? Parse is quite generic, more in line to what I'd name the filter-handler in stripes-components. But here, there might be `generateOptionalBooleanFilter` or `buildDateRangeFilter` etc.

mkuklis: @zburke I completely missed this comment but I agree. We should rename it!

PR #959, opened 2020-03-19
UIIN-921: Rewrite accordion controls

Baroquem:
https://issues.folio.org/browse/UIIN-921

Started as a small bug fix in expanding accordions. Ended as a complete rewrite of the accordion control code.

Veni, vidi, decimavi.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change history for ui-inventory
 
+## 2.1.0 (IN PROGRESS)
+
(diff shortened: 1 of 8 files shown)


Baroquem:
> Agreed, much nicer! Lemme run the integration suite past this branch before we merge so that if any changes are necessary there (because integration tests are brittle and sometimes lean harder than they should on HTML markup) then we can merge both sets of changes together.

Hi, @zburke ... let me know when you're ready!

PR #1002, opened 2020-04-20
UIIN-1007: Display instance updated date with instance status

Baroquem:
https://issues.folio.org/browse/UIIN-1007

This is a small change that displays the update date with instance status instead of in a separate field.


src/ViewInstance.js
```diff
@@ -741,17 +738,8 @@ class ViewInstance extends React.Component {
...
-            <FormattedDate value={instanceData.instanceStatusUpdatedDate} />
-          </p>
-          <p>
-            <FormattedTime value={instanceData.instanceStatusUpdatedDate} />
-          </p>
-        </Fragment>
-      )
+    const formattedStatusUpdatedDate = instanceData?.instanceStatusUpdatedDate !== '-'
+      ? getDateWithTime(instanceData.instanceStatusUpdatedDate)
       : noValue;
```
SergiyVSergiyenko, line 743: I am confused by the variety of `'-'` values. It is present here in its pure form, in the `noValue` value, and also the `getDateWithTime` function can return it.
What do you think if we make changes to your line 625 and here:
`const formattedStatusUpdatedDate = getDateWithTime(instanceData.instanceStatusUpdatedDate)` ?

Baroquem: Agreed. Thanks for your careful attention to detail!

PR #1023, opened 2020-05-07
UIIN-1036: A radical attempt to make the elements to be wrapped only …

SergiyVSergiyenko:
### Purpose
Make the elements to be wrapped only with a space symbol. Story [UIIN-1036](https://issues.folio.org/browse/UIIN-1036).
Continued #1010 and #1014.

### Screenshot
Screen Shot 2020-05-07 at 22 58 57


zburke:
Or do you want CSS like 
```
.no-wrap {
  white-space: nowrap;
}
```
and code like
```
return result.map(value => <span className={css.noWrap}>{value}</span>).join(', '); 
```
There's nothing actually telling the browser "don't wrap this, even though you might think these hyphens should allow you to do that."

PR #1043, opened 2020-05-28
UIIN-657 Item record: "loan type" appears as "loantype" (on detailed view)

K-Felk:
A number of the translation files had the "loan type" labels incorrectly entered 
as "loantype." I did a search for the incorrect label across the entire module and 
corrected it in every translation file in which it appeared.  

Refs: https://issues.folio.org/browse/UIIN-657


mkuklis:
@K-Felk thanks for doing this. The only file you will need to change is `en.json`. The rest of them will be built automatically.

K-Felk:
Ugh.  in that case, it's probably easier for me to just close this PR, delete the branch, and then recreate and apply the change where it's supposed to go, rather than to try and change it back in all the files.

PR #1056, opened 2020-06-12
UIIN-1137 make permissions localizable

zburke:
Refs [UIIN-1337](https://issues.folio.org/browse/UIIN-1137)


mkuklis:
Thanks for this @zburke. I didn't know how the permission keys should be defined. It's nice to see how this should be done.

PR #1072, opened 2020-06-19
UIIN-1158 Update Requests to SRS for v4

IvanKryzhanovskyi:
## Description
Request to SRS for formatted record changes

## Ticket
[UIIN-1158](https://issues.folio.org/browse/UIIN-1158)


IvanKryzhanovskyi:
> Adding SRS to the `package.json` `okapiDependencies` requires a bump to the app's major version on line 3.
> 
> Note also that if this is intended to be included in the 2020-Q2 release, in addition to merging this PR to master, you will also need to make a `b4.0` branch off the `b3.0` branch and merge these changes to that one as well. Please take care when doing this so we don't pollute the `b4.0` branch with other changes from `#master` that are not intended for Q2 release.

sure, I'll create separate MR for release branch

PR #1090, opened 2020-06-27
UIIN-829: Make language lists localizable

Baroquem:
https://issues.folio.org/browse/UIIN-829

## Purpose
A list of languages is used in inventory as a filter, as an instance record property, and to display the same property in the instance record view. Like most other text, these language name strings should be localizable.

## Approach
`react-intl` provides a function, `formatDisplayName`, which can return a localized version of a language name for its two- or three-character code. This is used to translate the language name labels in the aforementioned parts of ui-inventory. The catch is that `formatDisplayName` only returns the language name for a three-character code if there is no two-character code as well. Thus, both codes have to be considered, and the English name is used as a fallback if `formatDisplayName` fails to cooperate altogether.

The ISO 639-2 language/code list itself has been relocated to `stripes-components` for future cross-app utility (see https://github.com/folio-org/stripes-components/pull/1319/files). 

## Learning
The ISO 639 standard provides several canonical lists of languages and associated codes under several sub-standards (?): ISO 639-1 defines two-character codes (e.g., `la` for Latin); ISO 639-2 provides a much longer list of languages and language groups, and gives them three-letter codes (`lat` for Latin). There's also ISO 639-3, but for our purposes ISO 639-2 is most relevant. It's intended in part for bibliographic applications. The Library of Congress summarizes all this and offers script-friendly files of the ISO 639 language lists at https://www.loc.gov/standards/iso639-2/langhome.html.

Screen Shot 2020-06-27 at 11 47 20 AM
Screen Shot 2020-06-27 at 11 48 09 AM
Screen Shot 2020-06-27 at 11 48 32 AM
Screen Shot 2020-06-27 at 11 48 58 AM


src/Instance/InstanceDetails/InstanceDescriptiveView/utils.js
```diff
@@ -1,8 +1,19 @@
...
 // eslint-disable-next-line
 export const formatLanguages = (languages = []) => {
+  const intl = useIntl();
   return languages
-    .map(languageCode => languagetable.languageByCode(languageCode))
+    .map(languageCode => {
+      const intlCode = intlPreferredLanguageCode(languageCode);
+      // If formatDisplayName returns a localized language name, great -- use that.
+      // Otherwise, fall back to the English name.
+      return intl.formatDisplayName(intlCode, { fallback: 'none' }) || languageByCode(languageCode).name;
```
zburke, line 16: This same logic shows up here and in `InstanceFilters`. Is it worth wrapping into a function and exporting along with the other new functions you added in stripes-components?

mkuklis: I was not aware of `intl.formatDisplayName`. Nice work @Baroquem. I Agree with @zburke maybe this belongs to `stripes-components`?

Baroquem: @zburke @mkuklis You're both right, of course. I've done some cleanup and tried to simplify the whole setup. Michal, I wasn't aware of formatDisplayName either; 'twas Zak who pointed me in that direction!

PR #1096, opened 2020-07-02
UIIN-1112 Action menu view source in split screen instance records

NikitaSedyx:
## Purpose

https://issues.folio.org/browse/UIIN-1112

## Approach

View Source screen is moved to separate route to navigate to it without instance list-details opening
Move Slit screen items details to separate component to avoid copy paste of actions 

## Pre-Merge Checklist


src/Instance/InstanceMarc/InstanceMarc.css
```diff
@@ -0,0 +1,10 @@
+.instanceMarcWrapper {
+  margin-left: 20px;
```
aliaksei-chumakou, line 2: isn't there a constant for such paddings?

src/Instance/InstanceMarc/InstanceMarc.js
```diff
@@ -0,0 +1,78 @@
...
+        onClose={onClose}
+        data-test-instance-marc
+      >
+        <section className={styles.instanceMarcWrapper}>
+          <Headline
+            size="large"
+            margin="small"
+            tag="h3"
+          >
+            {intl.formatMessage({ id: 'ui-inventory.marcSourceRecord' })}
```
aliaksei-chumakou, line 44: so it's a standard now - to use formatMessage instead <FormattedMessage /> ?

src/Instance/InstanceMarc/InstanceMarc.css
NikitaSedyx: I'm not sure, but it was before in inline styles

src/Instance/InstanceMarc/InstanceMarc.js
NikitaSedyx: no, it's ok to use FormattedMessage - initially I had to use `intl.formatMessage` and as a result both could be used 

no difference for this case

test/bigtest/interactors/instance-marc-page.js
```diff
@@ -0,0 +1,21 @@
...
+  static defaultScope = '[data-test-instance-marc]';
+
+  isLoaded = isPresent('[data-test-pane-header-title]');
+
+  whenLoaded() {
+    return this.timeout(10000).when(() => this.isLoaded);
+  }
+
+  fields = collection('[data-test-instance-marc-field]');
+  close = clickable('[data-test-pane-header-dismiss-button]');
```
VictorSoroka1, line 18:
```suggestion
  closeButton = clickable('[data-test-pane-header-dismiss-button]');
```

src/Instance/InstanceMarc/InstanceMarc.js
VictorSoroka1: So I would suggest using FormattedMessage then from the general recommendation standpoint.

src/Instance/InstanceMovement/InstanceMovement.js
```diff
@@ -1,82 +1,41 @@
...
-            referenceData={referenceData}
-            draggable={false}
-            droppable={false}
-          />
-        </MoveItemsContext>
-      </InstanceDetails>
+      />
 
-      <InstanceDetails
+      <InstanceMovementDetailsContainer
```
VictorSoroka1, line 23: I think having defaults 
```
  instanceFrom: {},
  instanceTo: {},
```
 for the `<InstanceMovement />` component is not possible as it would lead to errors if the component would be instantiated with them. I suggest removing these defaults and mark them as required.

aliaksei-chumakou: Actually, it would be nice to have this as a rule of thumb - to not have default values as empty objects or arrays

test/bigtest/interactors/instance-marc-page.js
NikitaSedyx: it's an action

src/Instance/InstanceMarc/InstanceMarc.js
NikitaSedyx: what is general recommendation? :)

for me `intl.formatMessage` is better is it return just a string

src/Instance/InstanceMovement/InstanceMovement.js
NikitaSedyx: I'm pretty sure it wouldn't lead to errors when these props are not passed

aliaksei-chumakou: It's a good thing that you're sure, but I can refer to MCL, that makes thing difficult if you pass default empty array before your data is loaded - cell size measure will not work correctly.

src/Instance/InstanceMarc/InstanceMarc.js
```diff
@@ -0,0 +1,78 @@
...
+import styles from './InstanceMarc.css';
+
+const InstanceMarc = ({
+  instance,
+  marc,
+  onClose,
+}) => {
+  const intl = useIntl();
+
+  const renderFields = useMemo(() => ([
```
mkuklis, line 24: @NikitaSedyx could you do this processing in: https://github.com/folio-org/ui-inventory/pull/1096/files#diff-17145ff2fb1078fd7b5f31197e808f2cR31 right after you get the response back. Then you wouldn't need to use `useMemo` and actually not even pass the entire `marc` prop to `InstanceMarc` just the fields.

src/Instance/InstanceMovement/InstanceMovement.js
VictorSoroka1: So try it and we will see if you are sure.

src/Instance/InstanceMarc/InstanceMarc.js
VictorSoroka1: General recommendation is to use `FormattedMessage`.

src/Instance/InstanceMovement/InstanceMovement.js
VictorSoroka1: Screen Shot 2020-07-02 at 21 29 47
Screen Shot 2020-07-02 at 21 29 12

NikitaSedyx: ah, this error is thrown not by `InstanceMovementDetailsContainer`

> Anyways, what is the purpose of allowing to render the content if it is an empty object? It would lead to displaying of the content which is not quite ready or would possibly end up in error. I tracked down that everything is expected to work with the empty object which does not really make sense.

e.g. there is an error with data loading

options here:
for all show toast with "Smth is wrong"
1. to show nothing, empty pane, and with it props could be required
2. to show `the content which is not quite ready`
3. or actually show an error in the pane

@aliaksei-chumakou it's not the case with MCL, previously we discussed how to display pane when smth is wrong - content should be visible to end user # 2, but I agree about MCL (and IMO it should do recalculations when contentData is changed, but it's another story)

@VictorSoroka1 thanks for pointing possible error - it's holding loading

mkuklis:
@NikitaSedyx nice work as always!

src/Instance/InstanceMarc/InstanceMarc.js
NikitaSedyx: I'm ok to change to `FormattedMessage`, but could you point where it's recommended to use it?

VictorSoroka1: I remember [PRs though in the past](https://github.com/folio-org/ui-eholdings/pull/612) to use FormattedMessage when possible. Maybe @zburke or @JohnC-80 has more knowledge on this or maybe it is not actual now at all.

src/Instance/InstanceMovement/InstanceMovement.js
VictorSoroka1: Options that you listed I think is worth asking the PO. In data-export I prefer to show error notifications, but we don't have so many endpoints there. Also, I see that many apps are using the in build error alerts (e.g. `throwErrors: true`) which does not bring good UI experience but at least states that the error occurs. Don't have a strong opinion on that. My initial comment was about the possible errors because of the default props.

src/Instance/InstanceMarc/InstanceMarc.js
NikitaSedyx: I think it's because of `injectIntl`, but now we can use hooks

VictorSoroka1: Yeah, it can be true.

VictorSoroka1: There is also a [recommendation](
https://github.com/formatjs/formatjs/pull/1312/files#diff-96cc378dcdef180641176501f42183c7R98) from the react-intl devs.

> These should be used when your React component needs to format data to a string value where a React element is not suitable; e.g., a `title` or `aria` attribute, or for side-effect in `componentDidMount`.

zburke: In `react-intl` `v2`, the `injectIntl` implementation relied on React's deprecated context implementation and this is what led to our "Prefer `<FormattedMessage>`" recommendation. With `react-intl` `v4` the `injectIntl` HOC uses the modern context API, so `intl.formatMessage()` (whether you get to `intl` via `injectIntl` or the `useIntl` hook) is just as good as `<FormattedMessage>`. The [stripes i18n documentation](https://github.com/folio-org/stripes/blob/master/doc/i18n.md) reflects this now, but I didn't do a good job publicizing this change. 

And in any case, @NikitaSedyx is correct that it was always necessary to use `formatMessage()` when you needed a string-value.

VictorSoroka1: But in this particular case, as it turned out, the string representation is not necessary, so FormattedMessage can be used instead.

NikitaSedyx: my initial comment
> no, it's ok to use FormattedMessage - initially I had to use intl.formatMessage and as a result both could be used
no difference for this case

PR #1134, opened 2020-08-13
UIIN 907: Fast add record. Access New fast add template from Inventory	

K-Felk:
I'm able to add the button and get the fast add modal to show, but the
button does not visually match the other drop-down items.  The pluggable
component seems to have no way to tell the plug-in how it should
present itself.  I looked for a place off the menu to place this, but
the entire screen is a smart component, so I can't just place UI
elements anywhere I'd like.

In short, it works, but it looks funky, and I can't see how to
do anything about that.

refs: https://issues.folio.org/browse/UIIN-907


mkuklis:
@K-Felk unfortunately this is a bit tricky to handle not just because we need to be able to control plugin's button look but also because the plugin is embedded in the menu. When the `Pluggable` is loaded there will be no way for us to close the menu because it would also destroy the modal.

I made couple small adjustments inside the plugin (https://github.com/folio-org/ui-plugin-create-inventory-records/pull/11) so it should be now possible to embed it directly in `render` of `InstancesList` and control its visibility with the `open` props.

The setup should look similar to:

````js
<Pluggable
  buttonVisible={false}
  open={this.state.showNewFastAddModal}
  type="create-inventory-records"
  id="clickable-create-inventory-records"
  onClose={this.toggleNewFastAddModal}
/>
````

Then from the menu you could add something similar to:

````js
{this.getActionItem({
  id: 'new-fast-add-record',
  icon: 'plus-sign',
  messageId: 'ui-inventory.newFastAddRecord',
  onClickHandler: buildOnClickHandler(this.toggleNewFastAddModal),
})
````

Please let me know if this makes sense.

**Comment by K-Felk** [comment]
I inserted the pluggable component into the return value of the render, created a state variable to control the menu visibility, a simple toggle function to manage it, placed code in the drop-down to render the option, and made a new label in the translation file for it.  I then rebuilt my local environment with `yarn --force` to hopefully pull in the updated version of the plugin.  But it doesn't seem to be working-the "new fast add" button is showing, and the modal does not appear when the drop-down option is clicked.  I've pushed my changes here-am I doing something wrong?

**Comment by mkuklis** [comment]
@K-Felk all looks good to me! Perhaps try to remove `node_modules` and `yarn.lock` and run `yarn` again. It looks like perhaps the older version of the plugin still gets in for some reason.

**Comment by K-Felk** [comment]
I tried completely deleting the local repo, re-cloning it, and re-building from scratch.  Still borked.  :(

**Comment by mkuklis** [comment]
Very strange! What version of the plugin do you seen in your `yarn.lock`. It looks like we need: `1.0.10009`:

https://repository.folio.org/repository/npm-folioci/@folio/plugin-create-inventory-records/-/plugin-create-inventory-records-1.0.10009.tgz

**Comment by mkuklis** [comment]
Actually never mind. I just check that most recent version in the repo and it does not include the most recent changes. It looks like the newest version hasn't been publish. Very strange. I'm investigating this right now.

**Comment by mkuklis** [comment]
@K-Felk it looks like the version we want is `1.0.100012`. I do see it being published: 

https://repository.folio.org/#browse/browse:npm-folioci:%40folio%2Fplugin-create-inventory-records%2Fplugin-create-inventory-records-1.0.100012.tgz

Please check your yarn.lock. I'm curious what version do you see there.

**Comment by K-Felk** [comment]
Looks like it's pulling in 1.0.10009, so definitely a bit behind.

**Comment by K-Felk** [comment]
Not sure why, the package.json specifies "@folio/plugin-create-inventory-records": "^1.0.0", and my understanding is that a carat-range should allow for minor updates.

**Comment by mkuklis** [comment]
@K-Felk I wonder if it's some sort of yarn cache. Maybe `yarn cache clean` would help.

**Comment by K-Felk** [comment]
Tried that, it's still not pulling in the newest version.  And really, a complete rebuild should have cleared that as well, and doesn't seem to have worked.  @SergiyVSergiyenko and I both had problems getting the plugin to load at all locally-what we had to do was build ui-inventory by itself and add the plugin not only to the package.json file, but also put a reference in the local stripes config file.  It won't load if you just list it on the package.json, even if you clone the plugin code into the local workspace.

**Comment by K-Felk** [comment]
Okay, finally figured it out-there was a workspace-level cache that was overriding the local one-when I cleared the workspace, the new version finally started loading.  however, the modal still isn't showing when the drop-down option is clicked-instead, a vertical blue bar appears above the action menu.  I'll try moving the plugin placement-maybe there's a style that's keeping it from showing in the current location.

File: src/components/InstancesList/InstancesList.js [hunk 1]
```diff
@@ -213,6 +214,12 @@ class InstancesView extends React.Component {
     exportStringToCSV(cqlQuery);
   }
 
+  toggleNewFastAddModal() {
```
mkuklis, line 217: @K-Felk you may need to change the signature of this function to `toggleNewFastAddModal  = () => {` otherwise the `this` context is I think lost because of the way the function is invoked down below.

mkuklis:
@K-Felk that's great. I left you a little comment in the code. Maybe that will help.

src/components/InstancesList/InstancesList.js
K-Felk: Yup, good catch-changing the syntax got the modal to open.  I'll re-run tests and linter and then merge if it's clear.  Thanks for your help!

PR #1135, opened 2020-08-17
UIIN-1215: Add public and staff notes to `Holdings statements`

SergiyVSergiyenko:
https://issues.folio.org/browse/UIIN-1215

### Purpose
Add the ability to distinguish holdings statement notes as public or staff ones.

### Approach
Relabel the existing data element (note) as public, and add a new one marked as for staff.


src/constants.js
```diff
@@ -83,6 +83,21 @@ export const hridSettingsSections = [
...
+  },
+  {
+    type: 'holdingsStatementForSupplements',
+    title: 'Holdings statement for supplements',
+  },
+  {
+    type: 'holdingsStatementForIndexes',
+    title: 'Holdings statement for indexes',
+  },
+];
```
zburke, line 99: Shouldn't these `title` values be translation keys like `ui-inventory.holdings.statement` etc?

SergiyVSergiyenko: We use the value `title` only as a key in objects of the corresponding `Holdings statement` type. In other words, we don't use its value as a translation key for output to the UI.

Does it make sense in this case to use the `intl` library?

zburke: Ah, OK. I misunderstood how `title` was being used.

PR #1145, opened 2020-08-26
Hide an empty list when there are no items in `Holdings` record

SergiyVSergiyenko:
https://issues.folio.org/browse/UIIN-1076

## Purpose
Do not show an empty list if there are no items in the given `Holdings` record, and the draggable mode is deactivated.

### Demo

Hide_empty_list

### DIFF
--- a/src/Instance/HoldingsList/Holding/Holding.js
+++ b/src/Instance/HoldingsList/Holding/Holding.js
@@ -2,6 +2,7 @@ import React, {
   useMemo,
   useCallback,
   useContext,
+  useState,
 } from 'react';
 import PropTypes from 'prop-types';
 import { FormattedMessage } from 'react-intl';
@@ -50,6 +51,10 @@ const Holding = ({
   const { locationsById } = referenceData;
   const labelLocation = holding.permanentLocationId ? locationsById[holding.permanentLocationId].name : '';
   const withMoveDropdown = draggable || isDraggable;
+
+  const [open, setOpen] = useState(false);
+
+  const handleAccordionToggle = () => setOpen(!open);
 
   const viewHoldings = useCallback(() => {
     onViewHolding();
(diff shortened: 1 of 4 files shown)


src/Instance/HoldingsList/Holding/Holding.js
```diff
@@ -98,7 +101,8 @@ const Holding = ({
       }
       <Accordion
         id={holding.id}
-        closedByDefault={false}
+        open={open}
+        onToggle={() => setOpen(!open)}
```
NikitaSedyx, line 105: please don't create functions in props

src/Instance/ItemsList/ItemsListContainer.js
```diff
@@ -3,20 +3,24 @@ import React, {
...
 } from 'react';
 import PropTypes from 'prop-types';
+import { isEmpty } from 'lodash';
 
 import { stripesConnect } from '@folio/stripes/core';
 
 import ItemsList from './ItemsList';
 
-const ItemsListContainer = ({ holding, mutator, ...rest }) => {
+const ItemsListContainer = ({ holding, mutator, setOpen, ...rest }) => {
```
NikitaSedyx, line 12: it's not obvious from his component what `setOpen`  is for

src/Instance/HoldingsList/Holding/Holding.js
SergiyVSergiyenko: Thanks for pointing that out. Fixed.

src/Instance/ItemsList/ItemsListContainer.js
SergiyVSergiyenko: `setOpen` function controls the expanding/collapsing of the accordion depending on the presence/absence of elements in it.

PR #1151, opened 2020-09-03
UIIN-1211: Add settings page for fast add

SergiyVSergiyenko:
https://issues.folio.org/browse/UIIN-1211

## Purpose
Create the `Fast add` settings page with the ability to set `Instance status` and `Suppress from discovery` by default.

### Screenshot
Screen Shot 2020-09-03 at 12 13 59


Baroquem (approved):

mkuklis (approved):
@SergiyVSergiyenko this looks great. 

I maybe missing something but I was wondering about the code which actually saves the settings. Is that handled somewhere else already?

SergiyVSergiyenko:
You are absolutely right, @mkuklis. This was clearly not required in the original description. I returned the ticket to `In progress`.

mkuklis (approved):
This looks great @SergiyVSergiyenko. Thank you for adding the ability to save it.

PR #1159, opened 2020-09-17
UIIN-1259: Correct data validation in statistical code settings

K-Felk:
UI-inventory was not passing a validator function to the 
controlledVocabulary component, which was causing submission errors.
Those errors were confusing because they have very generic text, leading
to confusion about what was going wrong.  I re-wrote the validation function
to check that the statistical code was unique and hooked it up to the
component. Data is now being validated properly before being passed to the back end. 

Note that there are still some bugs in this particular settings pane that have
to do with CRUD operations-I have filed separate bug reports for those issues.

refs: https://issues.folio.org/browse/UIIN-1259

### DIFF
--- a/src/settings/StatisticalCodeSettings.js
+++ b/src/settings/StatisticalCodeSettings.js
@@ -46,9 +46,8 @@ class StatisticalCodeSettings extends React.Component {
 
     // if code has been entered, check to make sure the code value is unique
     if (item.code) {
-      const codes = items.map(({ code }) => code);
-      const count = codes.filter(x => x === item.code).length;
-      if (count > 1) {
+      const isDuplicated = items.some(({code}) => code === item.code);
+      if (isDuplicated) {
         errors.code = <FormattedMessage id="ui-inventory.uniqueCode" />;
       }
     }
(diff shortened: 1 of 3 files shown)


src/settings/StatisticalCodeSettings.js
```diff
@@ -41,11 +41,20 @@ class StatisticalCodeSettings extends React.Component {
...
 
-  validate = (item) => {
+  validate = (item, index, items) => {
     const errors = validateNameAndCode(item);
 
+    // if code has been entered, check to make sure the code value is unique
+    if (item.code) {
+      const codes = items.map(({ code }) => code);
+      const count = codes.filter(x => x === item.code).length;
+      if (count > 1) {
```
SergiyVSergiyenko, line 51:
```suggestion
      const isDuplicated = items.some(({code}) => code === item.code);
      if (isDuplicated) {
```

zburke (approved):
Hooray! I wonder how/when we introduced this terrible save --> "There are no [things]" bug??!?

SergiyVSergiyenko (approved):

PR #1161, opened 2020-09-18
UIIN-1268 error message on deleting a statistical code

K-Felk:
This was caused by an incorrect value for the 'shouldRefresh'
option in the data manifest.  That value is supposed to be a function,
not a simple boolean.  because of this, basically anytime the ControlledVocab
tried to do any write operation, it caused an error.  Changing the value to a simple
function that returns a boolean cleared up the problem.

refs:
https://issues.folio.org/browse/UIIN-1269
https://issues.folio.org/browse/UIIN-1268


zburke (approved):
Thanks for cleaning this up!

Minor style quibbles that should not hold up this PR:

* Add updates to the bottom of the list for the current release in the CHANGELOG. 
* In `shouldRefresh:() => true` you probably want a space after the colon; I'm surprised eslint doesn't complain about this honestly.

Baroquem (approved):

PR #1177, opened 2020-09-29
UIIN-1265: Move dictionary data fetching above router

mkuklis:
Move dictionary data fetching above the Router to avoid refetching the same data when switching between different routes.

https://issues.folio.org/browse/UIIN-1265


SergiyVSergiyenko (approved):

zburke (approved):
👏👏👏

PR #1186, opened 2020-10-06
UIDEXP-155: Implement instances search query saving to cql file instead of csv

YevheniiMaltsev:
## Purposes

Add implement instances search query saving to cql file instead of csv. I added a dependency 'file-saver' to cover file saving functionality. This dependency is already used in several modules (For example, ui-invoice) so the total bundle size should not be affected.

There is not way to check file downloading and its content using BigTest. So added a condition to prevent the files downloading during tests.

**Note:** it looks like this is a good idea to move the content saving to file to some of the core modules and keep it in one place. For now, there are several custom solutions in different places: ui-invoice, ui-data-export, ui-agreements. I suggest doing it after release and replace current custom solutions with a common one.

## Screenshots

Screenshot 2020-10-06 at 16 08 31
Screenshot 2020-10-06 at 16 08 42


mkuklis (approved):
Nice work @YevheniiMaltsev

zburke (approved):
I agree with @mkuklis, nice work! 

Thanks for the thoughtful comments in the PR description as well. It is good to know that our implementations of this type of feature are proliferating and that we should consider a single pattern/implementation. Good insight!

yelyzavetakhokhlova (approved):

PR #1212, opened 2020-11-05
UIIN-1324: Omit id from preceding/succeeding titles when duplicating instance

mkuklis:
Omit `id` from preceding/succeeding titles when duplicating instance

https://issues.folio.org/browse/UIIN-1324

### DIFF
--- a/src/components/InstancesList/InstancesList.js
+++ b/src/components/InstancesList/InstancesList.js
@@ -37,6 +37,7 @@ import {
   getCurrentFilters,
   parseFiltersToStr,
   marshalInstance,
+  omitFromArray,
 } from '../../utils';
 import {
   InTransitItemReport,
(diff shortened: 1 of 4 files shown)


src/components/InstancesList/InstancesList.js
```diff
@@ -129,9 +130,23 @@ class InstancesView extends React.Component {
...
+    }
+
+    if (succeedingTitles?.length) {
+      copiedInstance.succeedingTitles = omitFromArray(succeedingTitles, 'id');
+    }
+
     copiedInstance = set(copiedInstance, 'source', 'FOLIO');
-    this.setState({ copiedInstance });
+
+    if (instance?.precedingTitles) { this.setState({ copiedInstance }); }
```
zburke, line 149: Can you explain this conditional, @mkuklis? It really changes the logic of when `setState` is called and I don't understand (a) why it's changing or (b) why, if do in fact want the conditional, why we want it for `precedingTitles` only and not `succeedingTitles` also.

mkuklis: @zburke that's my mistake! Thank you for catching it.

mkuklis:
@zburke I think this should be ready now.

SergiyVSergiyenko (approved):

Baroquem (approved):

PR #1219, opened 2020-11-16
UIIN-1339: Return promise when submitting instance, holdings record or item form

mkuklis:
https://issues.folio.org/browse/UIIN-1339
https://issues.folio.org/browse/UIIN-1340

While submitting a form asynchronously FinalForm expects to receive a promise during submission in order to correctly set the `submitting` flag to `true`. 

https://final-form.org/docs/final-form/types/FormState#submitting

This PR is trying to address it.

The `submitting` flag was correctly used in ui-inventory in order to disable submit buttons during submission but the problem was that the flag was never set to `true` causing a "double submission" problem.


mkuklis:
I'm surprised to see that the test coverage for submission was not there before. Given this has to go out tomorrow I created a separate story for test coverage here: https://issues.folio.org/browse/UIIN-1351

SergiyVSergiyenko (approved):

Baroquem (approved):

PR #1220, opened 2020-11-17
UIIN-1349: Add select row checkboxes to inventory instance search result

ViktorSoroka07:
## Purpose

Add select row checkboxes to inventory instance search result in the scope of the [UIIN-1349](https://issues.folio.org/browse/UIIN-1349).

## Screenshots
Screen Shot 2020-11-17 at 1 56 17 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Do not escape quotes when submitting query-search queries. Refs UIIN-1319, UIIN-1320.
 * Use `TextArea` for some data fields on instance form. Refs UIIN-1278.
 * Return promise when submitting instance, holdings record or item form. Fixes UIIN-1339 and UIIN-1340.
+* Add select row checkboxes to inventory instance search result. Refs UIIN-1349.
 
 ## [5.0.1](https://github.com/folio-org/ui-inventory/tree/v5.0.1) (2020-10-15)
 [Full Changelog](https://github.com/folio-org/ui-inventory/compare/v5.0.0...v5.0.1)
--- a/src/components/InstancesList/InstancesList.js
+++ b/src/components/InstancesList/InstancesList.js
@@ -24,7 +24,8 @@ import {
 import { SearchAndSort } from '@folio/stripes/smart-components';
 import {
   Button,
-  Icon
+  Icon,
+  Checkbox,
 } from '@folio/stripes/components';
 
 import FilterNavigation from '../FilterNavigation';
@@ -45,6 +46,7 @@ import {
   InstancesIdReport,
 } from '../../reports';
 import ErrorModal from '../ErrorModal';
+import CheckboxColumn from './CheckboxColumn';
 
 import { buildQuery } from '../../routes/buildManifestObject';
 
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 * Add info callout when saving file with the instance UUIDs takes longer. Refs UIIN-1329.
 * Do not escape quotes when submitting query-search queries. Refs UIIN-1319, UIIN-1320.
 * Use `TextArea` for some data fields on instance form. Refs UIIN-1278.
+* Add select row checkboxes to inventory instance search result. UIIN-1349.
```
yelyzavetakhokhlova, line 14:
```suggestion
* Add select row checkboxes to inventory instance search result. Refs UIIN-1349.
```

yelyzavetakhokhlova (approved):

zburke (commented):
At present, we have a row-level click-handler for search results and the checkbox here interacts with that poorly: the click-target for the checkbox is very small, and clicking the background of the cell containing the checkbox activates the row's click-handler, while clicking the checkbox ignores it. 

Can/should we change this to put click-handlers on each cell instead so the `Titles` and `Contributors` cells are clickable but the checkbox cell, aside from the checkbox itself, is not? 

Thoughts, @JohnC-80?

ViktorSoroka07:
@zburke, the used approach is available already in folio, for example in data-import [here](https://folio-snapshot-load.dev.folio.org/settings/data-import/job-profiles/view?sort=name) and [licenses](https://folio-snapshot.dev.folio.org/licenses/) with the behavior you mentioned.

Assuming that we can end up with the changes you have proposed, I would suggest handling it in the scope of another story/PR.

zburke:
OK, that's fair, @ViktorSoroka07, but I feel like this UX needs review. I'll pursue that separately.

zburke (approved):

ViktorSoroka07:
Makes sense @zburke, thank you for the feedback.

src/components/InstancesList/InstancesList.js
```diff
@@ -365,6 +367,15 @@ class InstancesView extends React.Component {
...
 
     const resultsFormatter = {
+      'select': () => (
+        <FormattedMessage id="ui-inventory.instances.rows.select">
+          {ariaLabel => (
+            <CheckboxColumn>
+              <Checkbox aria-label={ariaLabel} />
+            </CheckboxColumn>
+          )}
+        </FormattedMessage>
```
zburke, line 377: `aria-label` needs to be a string but `<FormattedMessage>` with render-props provides an array in v5. Pull `intl { formatMessage }` out of `this.props` above then use 
```
<Checkbox aria-label={formatMessage({ id: 'ui-inventory.instances.rows.select' })} />

```

zburke (changes requested):
This looks good at a high level; nice work!

src/components/InstancesList/InstancesList.js
ViktorSoroka07: As I see in the browser the string is applied correctly even with the form I used. I think that in this case the array is converted to a string though.
```
<input id="checkbox-40" type="checkbox" aria-label="Select instance" class="input---2C-_P" aria-invalid="false" value="">
```
I saw the similar PR you made with the changes you suggested. Do you think this render prop should not be used at all?

zburke (approved):
Nicely done!

src/components/InstancesList/InstancesList.js
ViktorSoroka07: Per discussion with Zak, it is not recommended because of the prop-types warnings where `ariaLabel` is array, but `aria-label` expects string.

PR #1227, opened 2020-11-23
enable jest tests

aliaksei-chumakou:
## Pre-Merge Checklist


AndreiShumski2 (approved):

package.json
```diff
@@ -653,19 +653,31 @@
...
+    "@babel/plugin-proposal-decorators": "^7.0.0",
+    "@babel/plugin-transform-runtime": "^7.0.0",
+    "@babel/preset-env": "^7.0.0",
+    "@babel/preset-react": "^7.7.4",
     "@bigtest/interactor": "^0.7.0",
     "@bigtest/mocha": "^0.5.1",
     "@bigtest/react": "^0.1.2",
     "@folio/eslint-config-stripes": "^5.0.0",
     "@folio/stripes": "^5.0.0",
+    "@folio/stripes-acq-components": "^2.2.0",
```
ViktorSoroka07, line 672: I think more people should have reviewed this PR. I don't think ui-inventory should know about this package. This will make the development under the platform inconvenient. As a part of workspace all deps are popped up and thus running tests will result in errors in order to resolve which the deps should be setup outside the platform.

jest.config.js
```diff
@@ -0,0 +1 @@
+module.exports = require('@folio/stripes-acq-components/jest.config');
```
ViktorSoroka07, line 1: I don't think the idea of reusing the config from stripes-acq-components is good. As ui-inventory may end up with its own mocks and setup and in order to change them the changes would need to occur in that module which does not necessarily needed for stripes-acq-components. The first thing which I already found which differs between two modules is that babel-config should use legacy decorators as it is in many places the `@stripesConnect` notation is used.

This is how I did it in data-export.
```
  plugins: [
    [
      '@babel/plugin-proposal-decorators',
      { legacy: true },
    ],
    '@babel/plugin-proposal-class-properties',
    '@babel/plugin-transform-runtime',
  ],
```

package.json
aliaksei-chumakou: o_O Could you specify any implication or how it can make 

> This will make the development under the platform inconvenient
? I can't see any.
The rest is also unclear to me, what errors can it give?

jest.config.js
aliaksei-chumakou: I'm not sure that I've got your point. I've added initial setup for tests, any dev can change, fix or improve it. It works for me, and I'm not familiar with things you described, so I prefer to pass the buck of changing this. 
However, I can revert the PR if you think it's the best way to handle

ViktorSoroka07: No need to revert. I think you did well to set up it in a way it fulfills your need. I just think this can be improved to fulfill the needs of testing other more complex components. I think it deserves its own story.

package.json
ViktorSoroka07: I have this kind of failed tests when I install deps as a part of the platform. In order for this kind of errors to go away, I need to move the ui-inventory out of the platform and install all deps using yarn just for this only project so all deps ends up in the ui-inventory/node_modules as oppose to stripes/node_modules when we do it as a part of the platform. If you don't have such issues, could you please clarify in what way it solves for you?

aliaksei-chumakou: I don't develop module as a platform, it fails for me. I don't have such issues since tests - is a development thing, if you look at devDependencies - you'll find jest there. As well as stripes-acq-components dep.

jest.config.js
aliaksei-chumakou: sure, np. own story is nice

PR #1232, opened 2020-11-25
UIIN-1350: Implement records selection on the inventory search screen 

ViktorSoroka07:
## Purpose

Implement records selection on the inventory search screen in the scope of the [UIIN-1350](https://issues.folio.org/browse/UIIN-1350).

### DIFF
--- a/src/components/InstancesList/InstancesList.js
+++ b/src/components/InstancesList/InstancesList.js
@@ -90,6 +90,7 @@ class InstancesView extends React.Component {
     inTransitItemsExportInProgress: false,
     instancesIdExportInProgress: false,
     showErrorModal: false,
+    selectedRows: {},
   };
 
   onFilterChangeHandler = ({ name, values }) => {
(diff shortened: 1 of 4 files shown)


test/bigtest/tests/filters/instances/instances-list-test.js
```diff
@@ -49,5 +49,53 @@ describe('Instances list', () => {
...
+          await inventory.source.open();
+          await inventory.source.checkboxes.dataOptions(5).click();
+        });
+
+        it('should reset the selected state for the previously selected row', () => {
+          expect(instancesRoute.selectRowCheckboxes(0).isChecked).to.be.false;
+        });
+      });
+
+      describe('applying filters so previously selected items no longer displayed', () => {
```
yelyzavetakhokhlova, line 78:
```suggestion
      describe('applying filters so previously selected items are no longer displayed', () => {
```

src/components/InstancesList/InstancesList.js
```diff
@@ -348,6 +349,25 @@ class InstancesView extends React.Component {
     this.setState({ showErrorModal: false });
   };
 
+  toggleRowSelection = rowId => {
+    this.setState(({ selectedRows }) => {
+      const isRowSelected = Boolean(selectedRows[rowId]);
+      const newSelectedRows = { ...selectedRows };
+
+      if (isRowSelected) {
+        delete newSelectedRows[rowId];
```
yelyzavetakhokhlova, line 358: Do we need to delete that? I think it might look nicer if we drop the `if` statement by just setting row status to `false`.
 `newSelectedRows[rowId] = isRowSelected;`

ViktorSoroka07: In this case we will collect items which are not really needed. So at the end in order to get the selected items it would just enough to get object keys without any filtering logic.

yelyzavetakhokhlova: Makes sense.

yelyzavetakhokhlova (approved):

yelyzavetakhokhlova:
🚀

ViktorSoroka07:
By the way I committed your suggestion from GitHub. It was convenient from mobile🙂

yelyzavetakhokhlova:
> By the way I committed your suggestion from GitHub. It was convenient from mobile🙂

Yep, I saw that I became co-author. 😎 Working hard out there!

PR #1242, opened 2020-12-09
UIIN-1368: Update API endpoint for retrieving instances UUIDs

ViktorSoroka07:
## Purpose
Update API endpoint for retrieving instances UUIDs in the scope of the [UIIN-1368](https://issues.folio.org/browse/UIIN-1368).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 * Update title in Holdings detailed view. Fixes UIIN-1310.
 * Modify display of statistical codes in instance detail view. Refs UIIN-1150.
 * Result list. Align text in the columns in the top. Refs UIIN-1356.
+* Update API endpoint for retrieving instances UUIDs. Refs UIIN-1368.
 
 ## [5.0.1](https://github.com/folio-org/ui-inventory/tree/v5.0.1) (2020-10-15)
 [Full Changelog](https://github.com/folio-org/ui-inventory/compare/v5.0.0...v5.0.1)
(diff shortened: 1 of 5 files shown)


yelyzavetakhokhlova (approved):

KVupp (commented):
I think the permission re-name is missing, can you please change it in package.json

KVupp:
Also was just curious why the interface is not listed under okapi Interfaces? https://github.com/folio-org/ui-inventory/blob/master/package.json#L45?

ViktorSoroka07:
@KVupp, thank you. Handled permission renaming. Regarding the interface, since it works on the folio-snapshot it seems it is not needed in this case. Can it be so because some of the existing interfaces already include it?

ViktorSoroka07:
As I see "item-storage": "8.0" provides it.

https://github.com/folio-org/mod-inventory-storage/blob/fa058c357b9197da653b4d74ed5edf1a2f503ac5/descriptors/ModuleDescriptor-template.json

KVupp:
> As I see "item-storage": "8.0" provides it.
> 
> https://github.com/folio-org/mod-inventory-storage/blob/fa058c357b9197da653b4d74ed5edf1a2f503ac5/descriptors/ModuleDescriptor-template.json

item-storage itself is an interface Viktor, and instance-bulk is it's own interface. If it works on snapshot, it might not necessarily mean it is not needed, Atleast on the backend we use it to resolve dependencies. Probably better to confirm regarding the front end as well

ViktorSoroka07:
Gotcha, I have added the interface just in case.

KVupp:
> Gotcha, I have added the interface just in case.

Thanks Viktor!

PR #1246, opened 2020-12-11
UIIN-1373: Refactor the suppress actions into a single util function

SergiyVSergiyenko:
https://issues.folio.org/browse/UIIN-1373

## Purpose
As part of #1243, the `sourceSuppressor` util function was created. This PR eliminates duplication of a piece of code in `Settings` by using this function.


mkuklis (approved):
Very nice cleanup @SergiyVSergiyenko!

Baroquem (approved):

PR #1253, opened 2020-12-17
Support create inv plugin 2.0

doytch:
## Purpose
ui-plugin-create-inventory-records has a new major version v2 in development. [As part of that, I've added breaking changes in this PR.](https://github.com/folio-org/ui-plugin-create-inventory-records/pull/45) This PR handles those changes.

## Approach
- Remove IfPermission which is now being handled in the plugin.
- Switch to usage of `renderTrigger` instead of `buttonVisible` and a separate trigger render.


mkuklis (approved):
👏

PR #1268, opened 2021-01-11
UIIN-1393,UIIN-1395: Fix type ahead for location filters

mkuklis:
Fix type ahed for location filters:

https://issues.folio.org/browse/UIIN-1393
https://issues.folio.org/browse/UIIN-1395


SergiyVSergiyenko (approved):

SergiyVSergiyenko:
A simple elegant solution, @mkuklis! 👏

zburke (approved):
After reading through this a few times, I figured out the reason it works is a difference in the regular expressions between the default in stripes-components, `/^item.label/i`, and the one here in utils, `/item.[key]/` (no leading `^`). 

This is a bit more clear now that @mkuklis has added some documentation about how the different pieces work, but it's still not totally obvious that the `^` is really the biggest difference between the two.

PR #1270, opened 2021-01-12
Fix trivial ESLint errors

MikeTaylor:
We need to pick up our game on lint-policing: when "ignorable" errors begin to proliferate, significant errors get ignored. I fixed some quoting-and-spacing problems.

There still remains an error that mystifies me:
```
/Users/mike/ui-inventory/src/routes/InstancesRoute.test.js
  10:23  error  Unable to resolve path to module '@testing-library/user-event'  import/no-unresolved
```
Yet the package file specifies in the `devDependencies`:
```
    "@testing-library/user-event": "^12.1.10",
```
If anyone can see what's going wrong here, I'd like to know!

### DIFF
--- a/src/components/InstancesList/InstancesList.js
+++ b/src/components/InstancesList/InstancesList.js
@@ -573,7 +573,7 @@ class InstancesList extends React.Component {
             pagingType="click"
             hasNewButton={false}
             onResetAll={this.handleResetAll}
-            sortableColumns={["title","contributors","publishers"]}
+            sortableColumns={['title', 'contributors', 'publishers']}
           />
         </div>
         <ErrorModal


MikeTaylor:
BTW, there are also a _ton_ of `react-hooks/exhaustive-deps` lint errors, but I am not touching those as I don't feel confident that I understand hooks, let alone hook dependencies. Either someone who does understand should clean those us, or we should disable that rule in the `.eslintrc` file (which is what I have done locally).

doytch:
Agreed.

doytch (approved):

MikeTaylor:
@doytch Do you agree that someone needs to fix the `react-hooks/exhaustive-deps`, or that we need to disable that rule?

MikeTaylor:
(As discussed in a [Slack thread](https://folio-project.slack.com/archives/C210UCHQ9/p1610466959422600), Mark meant that someone should fix the `react-hooks/exhaustive-deps` errors ... although it's surprisingly difficult to find any explanation of that that rule wants you to do.)

PR #1298, opened 2021-02-09
UIIN-1412 parse item.enumeration numerically if possible

zburke:
Refactored item sort so that embedded numbers, e.g. `v23:n6`, are sorted by those numbers for enumeration, volume, and copyNumber. 

Refs [UIIN-1412](https://issues.folio.org/browse/UIIN-1412), [UIIN-1416](https://issues.folio.org/browse/UIIN-1416), [UIIN-1425](https://issues.folio.org/browse/UIIN-1425)


zburke:
I gave up on `_.sortBy` and just implemented real comparator functions.

src/Instance/ItemsList/utils.js
```diff
@@ -1,21 +1,53 @@
...
+  const aMatches = a.enumeration.match(/^v\.([0-9]+)(:no\.([0-9]+).*)?$/);
+  const bMatches = b.enumeration.match(/^v\.([0-9]+)(:no\.([0-9]+).*)?$/);
+  if (aMatches && bMatches) {
+    if (aMatches[1] === bMatches[1]) {
+      return Number.parseFloat(aMatches[3], 10) - Number.parseFloat(bMatches[3], 10);
+    } else {
+      return Number.parseFloat(aMatches[1], 10) - Number.parseFloat(bMatches[1], 10);
+    }
+  }
+
```
SergiyVSergiyenko, line 20: Indeed, it makes sense if there is a strict common pattern for `item.enumeration`. If not - 🤷‍♂️  .
In any case, we'd better wait for your question to be answered in the [comments](https://issues.folio.org/browse/UIIN-1412?focusedCommentId=97506&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-97506) to this ticket.

Baroquem (approved):
Seems like a reasonable approach to a problem that is always a source of headaches and that may not have a wholly elegant solution. As a point of interest and/or horror, I refer you to the [algorithm we previously used at Cornell](https://github.com/cul-it/blacklight-cornell-requests/blob/6ddc23b176226c3be74aa0ae2f7b3d9c3bb51192/app/models/blacklight_cornell_requests/request.rb#L289-L368) to handle this situation.

SergiyVSergiyenko (approved):

mkuklis (approved):
Nice work @zburke!

src/Instance/ItemsList/utils.js
```diff
@@ -1,21 +1,109 @@
...
+  'barcode': (a, b) => a.barcode.localeCompare(b.barcode),
+  'status': (a, b) => a.status.name.toLowerCase().localeCompare(b.status.name.toLowerCase()),
+  'copyNumber': (a, b) => a.copyNumber?.toLowerCase().localeCompare(b.copyNumber.name.toLowerCase()),
+  'materialType': (a, b) => a.materialType.name.toLowerCase().localeCompare(b.materialType.name.toLowerCase()),
+  'loanType': loanTypeComparator,
+  'effectiveLocation': (a, b) => a.effectiveLocation.name.toLowerCase().localeCompare(b.effectiveLocation.name.toLowerCase()),
+  'enumeration': enumerationComparator,
+  'chronology': (a, b) => a.chronology.localeCompare(b.chronology),
+  'volume': (a, b) => a.volume.localeCompare(b.volume),
+  'yearCaption': (a, b) => a.yearCaption.localeCompare(b.yearCaption),
```
mkuklis, line 102: I worry a bit that some of these fields are not required and may not be always present. Should we add `?.` to some of them like you did with `copyNumber`?

zburke: Thanks for pointing this out. I added `?.` where necessary and fixed the `yearCaption` comparator, which was outright broken because `yearCaption` is an array value, not a string value, and therefore can't directly use `localeCompare` without being converted to string first 🤦 .

mkuklis (approved):

zburke:
SonarCloud: fail
Me: SonarCloud, you keep on using this word. I do not think it means what you think it means. 

The local coverage report shows 100% coverage of statements, 91.3% of branches, 100% of functions, 100% of lines. Here, SonarCloud, try some iocaine. I'm sure you'll like it.

PR #1300, opened 2021-02-10
Eslint disable react hooks exhaustive deps

MikeTaylor:
At present, `yarn lint` on the master branch yields 44 errors, all of them `react-hooks/exhaustive-deps`. I have no idea what this means or how to fix it, and there is no online documentation of what this error describes.

Until someone comes along who is prepared to figure it out (perhaps @doytch at some point), these messages are pure noise which achieve nothing but obscuring real ESLint errors.


MikeTaylor:
I have no idea why this PR thinks it contains 64 commits. It really contains only one, and that is the addition of these three lines to the ESLint configuration:
```
  rules": {
    "react-hooks/exhaustive-deps": "off"
  }
```

doytch:
Mike, I don't know [what else to tell you](https://folio-project.slack.com/archives/C210UCHQ9/p1610472610430900?thread_ts=1610468081.426200&cid=C210UCHQ9). Did you find those explanations and examples of bugs that this lint rule avoids unsatisfactory?

doytch:
[Today I also found an RFC issue on React about the lint rules.](https://github.com/facebook/react/issues/14920#issuecomment-471070149) They got a bunch of examples of people who swore that the rule shouldn't apply to them, fixed some of the issues in later updates, and provided guidance on where it _does_ apply, and where it actually may be fine to eslint-ignore.

MikeTaylor:
It's still the case that not one of those three articles mentions `react-hooks/exhaustive-deps`, which doesn't give me confidence that this is something that community has embraced wholeheartedly. That, and the absence any documentation of the rule itself, as every other ESLint rule is documented, e.g. https://eslint.org/docs/rules/no-unneeded-ternary

Anyway ... I don't doubt that some of these warnings might be pointing to things that someone somewhere really ought to address. What I do doubt is that vomiting 44 giant errors across my screen every time I run ESLint is (A) likely to make me become that person, or (B) going to help me deal with lint errors relevant to my work.

So I want them gone. If you prefer for that to happen by fixing the errors, feel free :-)

doytch:
Mike, those links are literally pulled from the page you linked to as an example of something that had no info. The React documentation itself strongly recommends using the rules and they're written by the React team. The RFC issue says `react-hooks/exhaustive-deps` multiple times and has an absolute truckload of examples, guidance, and comments. What is the issue here? What am I missing?

I don't really feel like fixing this personally, when I was calling it out in real-time. Git-blame a few of the lines to figure out who added them, tag them in #stripes and ask them to fix it. After all, teams are supposed to have some percentage of time each sprint allocated to tech debt.

Here's a start: @NikitaSedyx and @aliaksei-chumakou

MikeTaylor:
It is not my wish to keep seeing all these meaningless (to me) errors in perpetuity until it suits someone else to make time to fix them — erspecially as the supposed errors do not impact in any way on my work.

But if I interpret you correctly, you are rejecting this PR. Is that correct? If so, let's make it explicit; then I can close it and go back to repeatedly `git stash`ing my fix to the ESLint config every time and make a commit and `git stash apply`ing back into place every time I need to run ESLint.

doytch:
If others actively want this rather than git-ignoring their local override or something, then fine. This isn't my module. I just know that if that happens this will never get fixed because that's how software development works, especially with a distributed team setup like we have on Folio. It's the reason it hasn't even been fixed by now.

But essentially, what you're saying is "we have a bunch of sketchy code here but rather than telling the people who wrote it, let's bury it under the bed." It's not even about the code, it's about the dangerous precedent this sets for a project that needs to enforce its coding guidelines via rules because all development is async and there is little shared culture.

I just didn't and continue to not understand what is so baffling about this rule. I've been asked for explanations and given them multiple times.

If somebody wants to tell me I'm actively wrong on these points, then go ahead and merge. Only thing I'd suggest is filing a JIRA against ui-inventory to actually fix the damn errors and remove the rule override so it doesn't get forgotten.

MikeTaylor:
Alright, let's close this and leave things as they are. If `yarn lint` is still reporting these errors in a year's time, I will return and taunt you, and you will have to buy me a beer.

(I'm not unsympathetic to your "if we stop nagging, it'll never get fixed" argument. I just don't see why _I_ should have have to be nagged about it all the time when it's nothing to do with me.)

mkuklis:
Thanks for closing this @MikeTaylor. 

I created: 

https://issues.folio.org/browse/UIIN-1415 to cover this. 

I'm not sure who will pick it up but I agree with everything @doytch said here.

aliaksei-chumakou:
@MikeTaylor are you talking about eslint errors or warnings? warnings are something for dev, errors should be fixed. Warnings for this rule you mentioned just to uncover for dev what some of vars are captured without dependency, and it's his responsibility to leave this or make things right. IMHO, we shouldn't put ignore-rule-comment on all these places, since you can hide mistake in a future, when new var will be used inside a hook, but won't be specified in deps array and no warning.
disabling the rule at all also is not an option.

PR #1317, opened 2021-02-23
UIIN-803: Add item counter to each holding record accordion

mkuklis:
Add item counter to each holdings record accordion. 

https://issues.folio.org/browse/UIIN-803

This took a bit longer because I had to move things around in order to get access to item records on the accordion level.

We could potentially improve this a bit more by fetching items for all holding records in one request.

holdings


zburke:
FYI I wrote to our friendly SonarCloud admin to ask about turning off its BS "Reliability Rating" since apparently it does not know how to evaluate `.?` 🙄

mkuklis:
Thank you @zburke! I tried to move it around but it was still complaining. So annoying.

Baroquem (approved):

SergiyVSergiyenko (approved):

PR #1318, opened 2021-02-24
UIIN-1406 Allow remote storage locations to be changed to non-remote storage location when moving the items between holdings

axelhunn:
https://issues.folio.org/browse/UIIN-1406
Allow remote storage locations to be changed to non-remote storage location when moving the items between holdings

## Pre-Merge Checklist


aliaksei-chumakou (approved):
interesting style of grouping modal parts... At least kind of unique

uladislausamets (approved):

mkuklis (approved):
Nice and clean @axelhunn. It's great to see react query in use too.

src/Instance/InstanceMovement/InstanceMovementContainer.js
```diff
@@ -39,15 +39,15 @@ const InstanceMovementContainer = ({
...
+    const instanceId = closedInstance.id === instanceFrom?.id
+      ? instanceTo?.id
+      : instanceFrom?.id;
 
     history.push({
       pathname: `/inventory/view/${instanceId}`,
       search: location.search,
     });
-  });
+  }, [history, location, instanceFrom, instanceTo]);
```
mkuklis, line 50: @axelhunn the `history` and `location` and `useCallback` doesn't use deep comparison

https://github.com/facebook/react/issues/14476

so this won't work as expected (it will basically recreate `onClose` more often). I would suggest either removing them from the dependency array or using something similar to:

https://github.com/kotarella1110/use-custom-compare

src/Instance/Move/useMoveItems.js
```diff
@@ -0,0 +1,57 @@
...
+    onSettled:  () => setIsMoving(false),
+    onError: error => {
+      const { message } = error;
+      callout.sendCallout({ type: 'error', message });
+    },
+    onSuccess: (_, variables) => {
+      const message = (
+        <FormattedMessage
+          id="ui-inventory.moveItems.instance.items.success"
+          values={{ count: variables?.itemIds.length }}
```
doytch, line 29: This just caught my eye because it looks incomplete. Either we know that `variables` and `itemIds` will always be there and we don't need the optional chaining at all or we need it completely with some nullish check.

```suggestion
          values={{ count: variables?.itemIds?.length ?? 0 }}
```

src/RemoteStorageService/index.js
```diff
@@ -0,0 +1,3 @@
+export * as Check from './Check';
```
doytch, line 1: I like the way you built up the chain for these components. They're clearly related and encoding that relationship in their access (eg, `RemoteStorage.Warning.ForHoldings`) hammers it into the readers head.

src/RemoteStorageService/Confirmation.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { Translate } from '../common';
+
+
+export const Heading = () => <Translate id="Removing from remote storage" />;
```
doytch, line 7: I _do_ like how this looks when browsing the code... 😆

src/RemoteStorageService/Check.js
```diff
@@ -0,0 +1,24 @@
+import { useRemoteStorageMappings } from '@folio/stripes/smart-components';
+import { useHoldings } from '../providers';
+
+
+export const useByLocation = () => {
+  const remoteMap = useRemoteStorageMappings();
+
+  return ({ fromLocationId, toLocationId }) => (fromLocationId in remoteMap) && !(toLocationId in remoteMap);
```
doytch, line 8: Could this be called with an `undefined` for either of the location ids? There're too many usages throughout the PR for me to track down completely, but it caught my eye. `undefined in remoteMap` will evaluate to `false`. The first thing that came to mind was situations where you could have some stale data or something.

It's legitimate to say that "no, that should never happen and it should spectacularly and loudly fail if it does." 😄

doytch (approved):
This looks good! I love the separation of smart/dumb work you've accomplished here with hooks. It looks easier to maintain and reason through than the previous mixing of business logic into components. Hopefully that bears out in the future!

And definitely a novel way of project structure compared to other spots in Folio but I see the good it brings.

I just put a couple of notes on minor things that I noticed.

src/RemoteStorageService/Confirmation.js
zburke: Yeah, but we batted it around at #stripes-architecture 2021-02-05 ago and _specifically decided against it_. 

I agree with you, @doytch, this does read nicely, but instead of ignoring established precedent and the decision we made as a group, @axelhunn please raise this again at #stripes-architecture and we can discuss the [existing tools that are built into `react-intl`](https://formatjs.io/docs/react-intl/components#message-descriptor) that provide the same functionality with `description` or `defaultMessage`? Maybe we can have the best of both worlds here.

src/RemoteStorageService/Check.js
```diff
@@ -0,0 +1,24 @@
...
+};
+
+
+export const useByHoldings = () => {
+  const { holdingsById } = useHoldings();
+  const check = useByLocation();
+
+  return ({ fromHoldingsId, toHoldingsId }) => {
```
mkuklis, line 16: @axelhunn I wonder if this should be wrapped  in `useCallback`. Otherwise I think it will be recreated on every render. What do you think?

src/RemoteStorageService/Check.js
axelhunn: The point here is to return `true` only if we move something from remote location to non-remote location.

- when `toLocationId` is `undefined` - it could be a black hole, but  it's not marked as remote, so
`(depends-on-from) && !(false)` === `depends-on-from`

- when `fromLocationId` is `undefined` - it certainly is not known remote location,
`(false) && !(whatever)` === `false`

- when both of them are `undefined` - the `fromLocationId` is undefined enough
`(false) && !(false)` === `false`

src/RemoteStorageService/Confirmation.js
axelhunn: @zburke sorry, didn't get it we decided against it.
Anyway, I made this part as more of a POC.
Reverted to unified way now, please review.

src/Instance/InstanceMovement/InstanceMovementContainer.js
axelhunn: Thanks for pointing this out. In fact I made this change for linter not to yell at me.
It happens the `history` is a mutable object (see https://reactrouter.com/web/api/history).
It is the same object every time, so we can rely on === comparison.

As for the `location` - it's not needed here, we can use `history.location` instead,

I've refactored the thing.

src/RemoteStorageService/Check.js
axelhunn: Since I've read this [article by Kent C. Dodds](https://kentcdodds.com/blog/usememo-and-usecallback), I try not to overuse the `useCallback`.

But you're right, this function is used in dependencies of another `useCallback`s and `useMemo`s.
Fixed it.

mkuklis: Thank you for sharing that article @axelhunn

PR #1321, opened 2021-02-25
Improved settings page for single-record-import profiles

MikeTaylor:
The old version simply used `<ControlledVocab>`, which is a real squeeze when you have nine editable fields. The new uses `<EntryManager>` and gives you 

I've retained the old version for now, as well as the new page, since we _know_ it does what we need, albeit in a really horrible way, and it's not impossible that there are surprises lurking in the brand new code. I'll rip out the old version once we have some experience with the new.

Fixes UIUN-1437, or at least constitutes _a_ fix. No doubt we will want to come back to this with some UX insight, and change the arrangement of the fields. But for now, we need something functional.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,7 @@
 * Inventory app now degrades gracefully in the absence of mod-copycat. Fixes UIIN-1428.
 * Modify language of Settings/Inventory/Single record import. Fixes UIIN-1429.
 * Disable the single-record-import modal's Import button until something has been entered. Fixes UIIN-1431.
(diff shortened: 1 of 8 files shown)


MikeTaylor:
All right, I am going to go ahead and merge. I don't have a formal review from @zburke but we have discussed on Slack, and having finally got the miraculous double-tick from the half-hour test runs, I am not risking someone else merging first and having to re-run them. (Not least, because doing so would take me past midnight.)

Zak, I hope the five recently added commits meet your hopes.

PR #1335, opened 2021-03-05
UIIN-1377: Add visual display when instance is suppressed from discovery

mkuklis:
This PR adds better visual indicators for suppressed instances. It takes care of two stories since they were very similar:

https://issues.folio.org/browse/UIIN-1377
https://issues.folio.org/browse/UIIN-1386

instances

### DIFF
--- a/src/Instance/InstanceDetails/InstanceDetails.js
+++ b/src/Instance/InstanceDetails/InstanceDetails.js
@@ -15,6 +15,7 @@ import {
   Pane,
   PaneMenu,
   Row,
+  TextField,
 } from '@folio/stripes/components';
 
 import { InstanceTitle } from './InstanceTitle';
@@ -30,6 +31,7 @@ import { InstanceClassificationView } from './InstanceClassificationView';
 import { InstanceRelationshipView } from './InstanceRelationshipView';
 import { InstanceNewHolding } from './InstanceNewHolding';
 import HelperApp from '../../components/HelperApp';
+import { WarningMessage } from '../../components';
 
 import {
   getAccordionState,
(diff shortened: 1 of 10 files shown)


src/Instance/InstanceDetails/InstanceDetails.js
```diff
@@ -119,11 +120,14 @@ const InstanceDetails = ({
...
-          <Row end="xs">
-            <Col
-              data-test-expand-all
-              xs
-            >
+          <Row>
+            <Col xs={10}>
+              {
+                instance.discoverySuppress &&
+                <TextField fullWidth readOnly value={intl.formatMessage({ id: 'ui-inventory.instance.suppressedFromDiscovery' })} error=" " />
```
SergiyVSergiyenko, line 127: Wouldn't it be better to put these attributes on multiple lines for better readability?

Baroquem (commented):
How does this address the case described in https://issues.folio.org/browse/UIIN-1386 of the instance having more than one warn-able status, e.g. the message "Instance is marked suppressed from discovery _and_ staff suppressed."?

mkuklis:
@Baroquem I followed the attached UX which has these messages separated.

src/Instance/InstanceDetails/InstanceDetails.js
SergiyVSergiyenko: Looks good to me! Thanks, @mkuklis  for making `WarningMessage` more reusable.

Baroquem:
Oddly, having previously left a comment, I can't see a button to approve this now. Approve!

PR #1346, opened 2021-03-15
Avoid error message from trying to fetch /copycat/profiles

MikeTaylor:
This is fetched in ImportRecordModal.js. That component was already,
rightly, not being included when the the `copycat-imports` interface
is unavailable. We now also omit it if the user doesn't have the
necessary permission, `copycat.profiles.collection.get`.

This permission is added as a sub-permission to
`ui-inventory.single-record-import`, since it's not possible to
perform an import without fetching a copycat profile.

Fixes UIIN-1447.

### DIFF
--- a/package.json
+++ b/package.json
@@ -716,6 +716,7 @@
         "displayName": "Inventory: Import single bibliographic records",
         "subPermissions": [
           "ui-inventory.instance.view",
+          "copycat.profiles.collection.get",
           "copycat.imports.post"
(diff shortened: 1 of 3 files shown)


MikeTaylor:
Crapola. Looks like yet another of our favourite random timeout failures. Except this one took a full hour.

Time to start the tests again and pray that no-one else successfully merges to master while it's happening.

PR #1347, opened 2021-03-16
UIIN-1451 Use the new contributorsNames index

hjiebsco:
Please see https://issues.folio.org/browse/RMB-797

## Purpose
To improve the contributor name search performance, a new index is added to backend. This change is to use that index.

## Approach
Change query template


zburke (approved):
@hjiebsco, will this index be available in R1? If yes, we can merge this now and include it in the R1 release. If not, we need to wait to merge this PR until after that release is published.

FYI, filed [UIIN-1451](https://issues.folio.org/browse/UIIN-1451) so we can track which ui-inventory release includes this improvement.

hjiebsco:
> UIIN-1451



> @hjiebsco, will this index be available in R1? If yes, we can merge this now and include it in the R1 release. If not, we need to wait to merge this PR until after that release is published.
> 
> FYI, filed [UIIN-1451](https://issues.folio.org/browse/UIIN-1451) so we can track which ui-inventory release includes this improvement.

Thanks @zburke. I would think this small fix can (should) be included in R1. @skoczko, can you confirm?

julianladisch (approved):
Looks good, thanks!

This changes from phrase match to word (in any order) match.
If there is a need the front-end may add support for a phrase match. This might be triggered by wrapping words into quotes.

mkuklis (approved):

PR #1354, opened 2021-03-23
UIIN-1044: Improve relationship with parent and child instances

mkuklis:
This PR addresses both UIIN-1044 and UIIN-1045:

https://issues.folio.org/browse/UIIN-1044
https://issues.folio.org/browse/UIIN-1045

- Add ability to pick an instance via `ui-plugin-find-instance`
- Add more info about parent / child sub instances

This may not function correctly because of the bug: 

https://issues.folio.org/browse/MODINV-389

parent-child

### Side notes: 

I had to turn off BigTest tests completely because they would never finish:

https://jenkins-aws.indexdata.com/job/folio-org/job/ui-inventory/job/UIIN-1044_and_UIIN-1045/

I created a JIRA to cover it https://issues.folio.org/browse/FOLIO-3092

@zburke reported that similar issues are happing in ui-users.

The coverage is below 80% because there are still a lot of missing tests we need to write in jest/RTL (for example form tests). The coverage for this work should be close to 100%.


src/Instance/InstanceDetails/SubInstanceList/SubInstanceList.js
```diff
@@ -0,0 +1,90 @@
...
+        to={`/inventory/view/${row[titleKey]}`}
+      >
+        {row.title}
+      </Link> :
+      row.title || '-'),
+    hrid: row => row.hrid || '-',
+    publisher: row => row.publication?.[0]?.publisher ?? '-',
+    publisherDate: row => row.publication?.[0]?.dateOfPublication ?? '-',
+    issn: row => getIdentifiers(row.identifiers, ISSN, identifierTypesById) || '-',
+    isbn: row => getIdentifiers(row.identifiers, ISBN, identifierTypesById) || '-',
```
SergiyVSergiyenko, line 41: I'm not totally sure but maybe in terms of `accessibility` it is better to use `noValue` from constants.js instead of '-'?

mkuklis: @SergiyVSergiyenko great idea. I was not aware of that!

mkuklis: We also have `<NoValue>` in stripes-components so I think I will go with that.

SergiyVSergiyenko (approved):

Baroquem (approved):

SergiyVSergiyenko:
Just a huge amount of work has been done: refactoring, custom hooks, etc.
Great job, @mkuklis! 👏

PR #1359, opened 2021-04-07
Configure maximum number of location to fetch in Stripes config

MikeTaylor:
It is now possible to configure the maximum number of location to
fetch in the Stripes config file, typically `stripes.config.js`, using
the `maxLocationsToLoad` entry in the `config` area.

Fixes UIIN-1480.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change history for ui-inventory
 
+## (IN PROGRESS)
+
+* It is possible to configure the maximum number of location to fetch in the Stripes config file, typically `stripes.config.js`, using the `maxLocationsToLoad` entry in the `config` area. Fixes UIIN-1480.
+
 ## [6.0.0](https://github.com/folio-org/ui-inventory/tree/v6.0.0) (2021-03-18)
 [Full Changelog](https://github.com/folio-org/ui-inventory/compare/v5.0.6...v6.0.0)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## (IN PROGRESS)
 
-* It is possible to configure the maximum number of location to fetch in the Stripes config file, typically `stripes.config.js`, using the `maxLocationsToLoad` entry in the `config` area. Fixes UIIN-1480.
+* It is possible to configure the maximum number of location to fetch in the Stripes config file, typically `stripes.config.js`, using the `maxUnpagedResourceCount` entry in the `config` area. Fixes UIIN-1480.
 
 ## [6.0.0](https://github.com/folio-org/ui-inventory/tree/v6.0.0) (2021-03-18)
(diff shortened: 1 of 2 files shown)


zburke (approved):
OMG this is so obvious in retrospect. The only question I have is whether `maxLocationsToLoad` should instead be something like `maxUnpagedResourceCount` (an "unpaged resource" being one where we are implying "Give them all to me in one list, I don't care how many there are"). That would let us avoid a proliferation of `maxFoo` and `maxBar` values. OTOH, maybe per-resource, er, per-resource-type settings are valuable. 

Related, how do we surface settings like these so implementors know they can use them in their `stripes.config.js` files? 

@doytch, @mkuklis, any thoughts on this pattern?

MikeTaylor:
> OMG this is so obvious in retrospect. The only question I have is whether `maxLocationsToLoad` should instead be something like `maxUnpagedResourceCount` (an "unpaged resource" being one where we are implying "Give them all to me in one list, I don't care how many there are"). That would let us avoid a proliferation of `maxFoo` and `maxBar` values. OTOH, maybe per-resource, er, per-resource-type settings are valuable.

Oh, yes, that's much better. I'll tweak accordingly.

MikeTaylor:
Now to wait another hour for the tests to re-run ...

mkuklis:
@MikeTaylor I like this approach. Very clever!  

I wish we introduce some sort of module namespace for that config. Something like `config.inventory.maxUnpagedResourceCount`. 

I think this is related to what @zburke is asking. Perhaps we should introduce a separate place for these kind of `customizations`?

MikeTaylor:
Well, it's open to debate whether we consider value we specify here to be a property of the Inventory app (in which case your namespacing idea is not bad) or of locations (in which case my original name was OK) or more generally of "How FOLIO should do this stuff" (as with the name I adopted that @zburke suggested.)

Maybe the full generality is something like:
```
config: {
  unpagedResourceCount: {
    default: 500,
    resources: {
      locations: 1000,
      copyrightStatuses: 20,
    },
    apps: {
      inventory: 800,
      users: 700,
    },
    appResources: {
      inventory: {
        locations: 2000,
      }
    } 
  }
}
```

But, really, who needs it? In practice this is just about avoiding an endless sequence of trivial "increase 1000 to 1500" Jiras which, because of all the bureaucracy around them, end up taking multiple days to handle.

mkuklis:
@MikeTaylor everything you said is valid. I was thinking that this kind of approach could be valuable for various other vendor specific customizations/additional config in the future. I wonder if we should consider introducing some sort of `customization.config` which could be handled by stripes. Just thinking here for now.

I think what you've done is prefect.

MikeTaylor:
I agree it would not be a bad thing to introduce a compilable config, to which Stripes would furnish a simple API call, `props.stripes.config('unpagedResourceCount', 'inventory', 'locations')` — the arguments being parameter-name, module-name and resource-name, where either or both of the last two can be omitted.

But that is for another day, I guess.

PR #1368, opened 2021-04-15
UIIN-1481: Fix date display on item view

mkuklis:
The dates on view and edit screens were out of sync because the missing `timeZone`.

https://issues.folio.org/browse/UIIN-1481


zburke (approved):
Alternatively, see [`<FormattedUTCDate>`](https://github.com/folio-org/stripes-components/tree/master/lib/FormattedUTCDate) which does literally the exact same thing. The only benefit would be that `<FormattedUTCDate>` has some documentation attached to explain why it works. 

I'm still a little confused by why `<FormattedDate value="2021-04-14">` would ever display as `2021-04-13` without the `timeZone` prop, but that's not a quarrel with your work. 

Thank you!

mkuklis:
Oh nice I was not aware of `<FormattedUTCDate>`. I'm happy to replace it.

PR #1378, opened 2021-04-26
STRIPES-741: Add <Pluggable> for checking copyright

AndrewIsh:
STRIPES-741

## Purpose

This commit adds a `<Pluggable>` to the main Instance action menu to
enable plugins that provide copyright checking functionality to add
themselves to the menu, via the `copyright-permissions-checker` plugin type.

This was added to enable the CLA Permissions Check plugin
(https://github.com/PTFS-Europe/ui-plugin-cla-permissions-check) to add
itself.

Additional state was added to the component to maintain the open status
of a modal that the plugin may present, this state is passed to the
plugin, along with a function to set it.

Since such plugins will require an identifier for the instance to enable
them to check copyright status, a new function getIdentifiers was added
which, when passed an instance's data, will extract a suitable
identifier to pass to the plugin

https://issues.folio.org/browse/STRIPES-741

## Approach

See above, this approach follows the established pattern of using the `<Pluggable>` to allow hooks into existing UI

## Pre-Merge Checklist
  - [ ] Code coverage on new code is 80% or greater (the file in question, ViewInstance.js currently has 0 coverage, it seems out of scope of this PR to give it coverage).

### DIFF
--- a/src/ViewInstance.js
+++ b/src/ViewInstance.js
@@ -309,26 +309,23 @@ class ViewInstance extends React.Component {
     if (!records) {
       return null;
     }
-    const allIdentifiers = [].concat(
-      records.map(record => record.identifiers)
-    )[0];
-    const result = [];
-    allIdentifiers.forEach(({ identifierTypeId, value }) => {
-      const ident = identifierTypesById[identifierTypeId];
-      if (
-        (ident?.name === ISBN ||
-        ident?.name === ISSN) &&
-        value
-      ) {
-        result.push({
-          type: ident.name,
-          value
-        });
-      }
-    });
     // We can't make any meaningful assessment of which is
     // the best identifier to return, so just return the first
-    return result.length > 0 ? result[0] : null;
+    // we find
+    for (const record of records) {
+      for (const identifiers of record.identifiers) {
+        const { identifierTypeId, value } = identifiers;
+        const ident = identifierTypesById[identifierTypeId];
+        if (
+          (ident?.name === ISBN ||
+            ident?.name === ISSN) &&
+          value
+        ) {
+          return { type: ident.name, value };
+        }
+      }
+    }
+    return null;
   };
 
   createActionMenuGetter = instance => ({ onToggle }) => {
--- a/src/ViewInstance.js
+++ b/src/ViewInstance.js
@@ -12,6 +12,7 @@ import {
   AppIcon,
   IfPermission,
   IfInterface,
+  Pluggable,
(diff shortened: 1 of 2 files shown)


AndrewIsh:
Hi all! What's the best way to move this forward? Does the PR need more work? Thanks!

src/ViewInstance.js
```diff
@@ -289,10 +293,44 @@ class ViewInstance extends React.Component {
...
 
+  // Get all identifiers for all records
+  getIdentifiers = (data) => {
+    const { identifierTypesById } = data;
+    const { ISBN, ISSN } = indentifierTypeNames;
+    const records = this.props?.resources?.selectedInstance?.records;
+    if (!records) {
+      return null;
+    }
+    const allIdentifiers = [].concat(
```
mkuklis, line 312: @AndrewIsh what's the purpose of `concat` here?

src/ViewInstance.js
```diff
@@ -289,10 +293,44 @@ class ViewInstance extends React.Component {
...
+      records.map(record => record.identifiers)
+    )[0];
```
mkuklis, line 314: @AndrewIsh it looks like only first record in the array is being used here. What if the first record has no identifiers at all? Sorry for this question but I'm having a hard time understanding what `getIdentifiers` does. The comment says `Get all identifiers for all records` but the code doesn't seem to do that. Could you please clarify what is happening here?

mkuklis (commented):
@AndrewIsh I left couple questions about `getIdentifiers` for you.

mkuklis:
@AndrewIsh I apologize for not getting back to you sooner on this. For some reason I was not notified by github about this PR.

src/ViewInstance.js
AndrewIsh: Arghh, good spot. I suspect I was originally deep cloning the array then switched to `map`, but forgot to remove the `[].concat()` I'll fix that.

src/ViewInstance.js
AndrewIsh: > I think selectedInstance will only return a single record right
That's my understanding. Whereas `records` is an array but only ever seems to contain a single element, but I'm assuming it can contain more.

AndrewIsh:
Also switched from `for` loop to `for....of` loop to address code smells

AndrewIsh:
> @AndrewIsh nice work! I'm going to merge this in if that's ok with you?

If you're happy with it, please do!

PR #1427, opened 2021-07-28
UIIN-1470: Update pane headers

Baroquem:
https://issues.folio.org/browse/UIIN-1470

This is a redo of the work previously merged as UIIN-1470; the spec in the Jira ticket turned out to be incorrect the first time around.

Screen Shot 2021-07-28 at 4 10 32 PM


zburke (approved):

NikitaSedyx:
hi @Baroquem seems this request produced a regression, Holding View page refresh throws and error `Error: TypeError: Cannot read property 'effectiveLocationId' of undefined`

https://github.com/folio-org/ui-inventory/pull/1427/files#diff-a962523d8df448cf50db88af841c0c0602e66f93913b469671082cea6cf85667R642

Baroquem:
Hi @NikitaSedyx ... this seems to work without errors for the sample records I've tested in snapshot and bugfest-juniper. Do you have an example of a record that causes the error that I could look at?

PR #1439, opened 2021-08-18
UIIN-1500: Add links to headers in item detail view and holdings detail view

Baroquem:
https://issues.folio.org/browse/UIIN-1500
https://issues.folio.org/browse/UIIN-1501

This PR adds links in the header of the item detail view to the instance record and holdings record, and a link in the header of the holdings detail view to the instance record.

Screen Shot 2021-08-18 at 2 24 25 PM
Screen Shot 2021-08-18 at 2 24 15 PM


src/views/ItemView.js
```diff
@@ -815,10 +815,10 @@ class ItemView extends React.Component {
 
               <Row center="xs">
                 <Col sm={6}>
-                  <FormattedMessage
-                    id="ui-inventory.instanceTitle"
-                    values={{ title: instance.title }}
-                  />
+                  <FormattedMessage id="ui-inventory.instanceTitle" />
+                  <Link to={`/inventory/view/${instance.id}`}>
+                    {` ${instance.title}. `}
```
mkuklis, line 820: @Baroquem I'm curious was there a reason to move the actual title out of `FormattedMessage`?

mkuklis (approved):

src/views/ItemView.js
Baroquem: I suppose it would be better not to, wouldn't it? Maybe something like this instead?
`    const linkedInstanceTitle = (
      <Link to={`/inventory/view/${instance.id}`}>
        {` ${instance.title}. `}
      </Link>
    );
`
`                <FormattedMessage
                    id="ui-inventory.instanceTitle"
                    values={{ title: linkedInstanceTitle }}
                  />
`

zburke (approved):

PR #1441, opened 2021-08-23
UIIN-1569 Make sure building inventory module works with babel-plugin-lodash

mkuklis:
https://issues.folio.org/browse/UIIN-1569

`babel-plugin-lodash` currently complains about lodash using chain sequences. 

This PR refactors it to use `flow`.

````
inventory/src/hooks/useLoadSubInstances.js: Lodash chain sequences are not supported by babel-plugin-lodash.
Consider substituting chain sequences with composition patterns.
See https://medium.com/making-internets/why-using-chain-is-a-mistake-9bc1f80d51ba
  22 |   const results = useInstancesQuery(instanceIds.map(inst => inst[subId]));
  23 |   const allLoaded = results.reduce((acc, { isSuccess }) => (isSuccess && acc), true);
> 24 |   const instances = chain(results)
````

More info:

https://medium.com/bootstart/why-using-chain-is-a-mistake-9bc1f80d51ba

We need this in place in order to finalize work on https://issues.folio.org/browse/STRWEB-20

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 * Fix instance duplication when child or parent records are present. Fixes UIIN-1562.
 * Add bound-with icons and item detail view header note. Refs UIIN-1522, UIIN-1523, UIIN-1524.
 * Retrieve and display select Piece information on Holding. Refs UIIN-1502.
-
+* Make sure building inventory module works with `babel-plugin-lodash`. Refs UIIN-1569.
 
 ## [7.1.2](https://github.com/folio-org/ui-inventory/tree/v7.1.2) (2021-07-14)
 [Full Changelog](https://github.com/folio-org/ui-inventory/compare/v7.1.1...v7.1.2)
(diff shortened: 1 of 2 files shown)


Baroquem (approved):

zburke (commented):
Can you explain why the native JS methods are insufficient here? i.e., if we need to refactor away from `chain`, why not just ditch lodash altogether? If this is a simpler refactor, that's cool. I just don't understand why lodash was involved in the first place.

zburke:
Answering my own question: `flow` allows you to [avoid using `partialRight`](https://stackoverflow.com/questions/49799024/better-and-more-performant-than-lodash-chain) if the function requires arguments in addition to the iteratee. ([lodash doc](https://lodash.com/docs/4.17.15#partialRight)).

zburke (approved):

mkuklis:
@zburke you are right we could just use pure js functions here. The `sortBy ` would get more verbose.

PR #1444, opened 2021-08-25
UIIN-1567: Incorporate ui-inventory-search facets

mkuklis:
Incorporate facets from `ui-inventory-es`.

https://issues.folio.org/browse/UIIN-1567

facets


zburke:
The conclusion at sprint planning today was that separating the migration from the test-writing is, indeed the way to do. I created [UIIN-1572](https://issues.folio.org/browse/UIIN-1572) as a top-level ticket that others can be linked to.

package.json
```diff
@@ -48,11 +48,12 @@
...
-      "inventory": "10.10",
+      "inventory": "10.10 11.0",
       "source-storage-records": "3.0",
-      "instance-storage": "7.0",
-      "holdings-storage": "3.0 4.4",
-      "item-storage": "8.0",
+      "search": "0.5",
+      "instance-storage": "7.0 8.0",
+      "holdings-storage": "3.0 4.4 5.0",
+      "item-storage": "8.0 9.0",
```
marcjohnson-kint, line 56: Why does using the `search` interface need changes to the other interface versions as well?

zburke: It doesn't; that was just an artifact of the merge with the main branch.

zburke (approved):
Filed UIIN-1572 for test coverage.

package.json
marcjohnson-kint: I don't understand. If the main branch already expressed the interface dependencies that way, a comparison between this branch and the main branch should not describe it as a change.

PR #1453, opened 2021-09-07
UIIN-1521: Add bound-with accordion to item view

Baroquem:
https://issues.folio.org/browse/UIIN-1521

This adds an accordion to the bottom of the item view page that displays the contents of the item record's `boundWithTitles` prop, if any.

Screen Shot 2021-09-07 at 4 31 16 PM

PR #1456, opened 2021-09-10
UIIN-1521: Add bound-with accordion to item view

Baroquem:
https://issues.folio.org/browse/UIIN-1521

This is an updated version of https://github.com/folio-org/ui-inventory/pull/1453 with the attempt to mock react-intl removed.


mkuklis (approved):
@Baroquem nice work! It's great to see the tests green again!

PR #1460, opened 2021-09-17
UIIN-1634 only search requests should use mod-search

zburke:
When we implemented ElasticSearch, all requests, instead of only `GET`
requests, were redirected to the mod-search endpoints. This restores the
mod-inventory endpoints for `POST`, `PUT`, and `DELETE` requests,
restoring the ability to create, edit, and remove records.

Refs [UIIN-1634](https://issues.folio.org/browse/UIIN-1634)


mkuklis:
@zburke nicely done!

PR #1468, opened 2021-09-28
UIIN-1639: Make mapped fields read-only for MARC holdings records

Vladyslav-Velytskyi:
# UIIN-1639: Edit FOLIO Inventory Holdings record > When Source = MARC > Then make mapped fields read-only

## Purpose
Add `disabled` attribute for necessary fields.

Issue: [UIIN-1639](https://issues.folio.org/browse/UIIN-1639)

Related [PR](https://github.com/folio-org/stripes-smart-components/pull/1130) in `stripes-smart-components`

## Pre-Merge Checklist

## Screenshots
3PYep5QqHC


BogdanDenis (approved):

mariiaschv (approved):

src/edit/electronicAccessFields.js
```diff
@@ -80,12 +81,14 @@ ElectronicAccessFields.propTypes = {
   canAdd: PropTypes.bool,
   canEdit: PropTypes.bool,
   canDelete: PropTypes.bool,
+  isMARCRecord: PropTypes.bool,
```
mkuklis, line 84: @Vladyslav-Velytskyi could we use `canEdit` for this instead of introducing a new prop?

mkuklis (approved):

zburke (changes requested):
MARC-ness and editability happen to overlap here, but they are not the same. As @mkuklis suggested, instead of checking `isMARCRecord` everywhere, use `isMARCRecord` to set `canEdit` and trigger `disabled` based on `canEdit` alone. I know it seems petty, but having centralized logic will be really nice for the next time when we only have to change 
```
isEditable = canEdit && !isMARCRecord && !hasOtherReadOnlyCondition;
```
in one place.

src/edit/electronicAccessFields.js
Vladyslav-Velytskyi: Thank you, nice suggestion. But here is one problem, `RepeatableField` has prop `canEdit` which doesn't work. It is thrown directly into a `FieldArray` that does not support `canEdit` prop. Ыo it was necessary to add support for this prop in `RepeatableField`. By the way, it looks better then with `isMARCSource`

PR #1469, opened 2021-09-28
UIIN-1603: Add Jest tests for languages, utils

Baroquem:
https://issues.folio.org/browse/UIIN-1603
https://issues.folio.org/browse/UIIN-1613

Test coverage for utils is incomplete, but it's taking a long time and there are higher-priority components to cover. I'm merging what I have so far, and we can circle back to finish it later.


mkuklis (approved):
It looks great @Baroquem! Thank you for the extra docs!

PR #1482, opened 2021-10-12
UIIN-1642: Fix impossibility to create `in transit` items report

Stanislau-Snapok:
## Purpose
Fix impossibility to create `in transit` items report.

## Approach
Now we can got all fields necessary for report from `inventory-reports/items-in-transit` end-point. The descriptions of fields here https://s3.amazonaws.com/foliodocs/api/mod-circulation/inventory-reports.html#inventory_reports_items_in_transit_get 
are matching with descriptions for fields in epic and stories related to `in transit` report https://issues.folio.org/browse/UXPROD-944.

The `inventory-reports/items-in-transit` end-point didn't takes any query parameters, so I guess it should give us all items which have status `in transit`.

Now we can generate report. One of them attached below.

Also I renamed `InTransitItemReport` into `InTransitItemsReport` in `InstancesList` and index file. In my mind as it should be.

## Refs
https://issues.folio.org/browse/UIIN-1642

## Attachments
[InTransit (8).csv](https://github.com/folio-org/ui-inventory/files/7330891/InTransit.8.csv)


src/reports/index.js
```diff
@@ -1,2 +1,2 @@
-export { default as InTransitItemReport } from './InTransitItemReport';
+export { default as InTransitItemsReport } from './InTransitItemReport';
```
mkuklis, line 1: @Stanislau-Snapok could you please rename the file to `InTransitItemsReport` as well?

PR #1494, opened 2021-10-20
UIIN-1651: Re-enable date filters

Baroquem:
https://issues.folio.org/browse/UIIN-1651
https://issues.folio.org/browse/UIIN-1653

This re-enables the instance date filters for date created and date updated, which were temporarily disabled during the transition to ElasticSearch.  It also suppresses errors from the /instances/facets endpoint.

Screen Shot 2021-10-20 at 4 44 12 PM


mkuklis (approved):
Nice work @Baroquem!

PR #1503, opened 2021-10-28
UIIN-1828 Create a MARC Holdings Record

mariiaschv:
## Purpose
Add support of editing MARC Holdings records.

Issue: [UIIN-1828](https://issues.folio.org/browse/UIIN-1828)

## Related PR
https://github.com/folio-org/ui-quick-marc/pull/222


src/ViewInstance.js
```diff
@@ -434,7 +440,7 @@ class ViewInstance extends React.Component {
           </IfInterface>
 
           {
-            canEditDeriveMARCRecord && (
+            (canCreateMARCHoldingsForInstanceWithSourceMARC || canEditDeriveMARCRecord) && (
```
BogdanDenis, line 443: We don't need to change anything with `View source` option because it's not related to Holdings. I think we should render it when `canEditDeriveMARCRecord` is `true` and that's it

src/ViewInstance.js
```diff
@@ -476,7 +477,7 @@ class ViewInstance extends React.Component {
         </MenuSection>
 
         {
-          canCreateEditDeriveMARCRecord && (
+          (canCreateMARCHoldingsForInstanceWithSourceMARC || canEditDeriveMARCRecord) && (
```
BogdanDenis, line 480: Same here for `Edit in quickMARC`

BogdanDenis (changes requested):

src/ViewInstance.js
mariiaschv: Yeah, you are right. Will fix it, thanks.

src/ViewInstance.js
mariiaschv: Hmm, but this condition is not for `Edit in quickMARC` but for showing quickMARC section in the Instance record Actions menu. It needs to be displayed when user can create MARC Holdings OR when user can edit/derive MARC bib. And then each option of this menu section will or will not be displayed according to user permissions. Correct me, if I'm wrong, please :)

BogdanDenis: Yes, you're right. I thought this was just for editing in quick marc

BogdanDenis (approved):

Vladyslav-Velytskyi (approved):

mkuklis (approved):

PR #1509, opened 2021-11-08
UIIN-1838: Fix successful toast appears after clicking on In transit items report (CSV)

Dmitriy-Litvinenko:
## Purpose
Fix successful toast appears after clicking on In transit items report (CSV)

## Approach
- Changed toast from info to success 
- Added missing translation

## Stories
https://issues.folio.org/browse/UIIN-1838

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 * Change search operators for ISBN and ISSN to '='. Fixes UIIN-1846.
 * Allow user to sort receiving history table. Refs UIIN-1824.
 * Change search operator for "Identifier (all)" index to '='. Fixes UIIN-1855.
+* Fix successful toast appears after clicking on `In transit items report (CSV)`. Refs UIIN-1838.
 
 ## [8.0.0](https://github.com/folio-org/ui-inventory/tree/v8.0.0) (2021-10-05)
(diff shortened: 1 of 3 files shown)


pavelspichonak (approved):
LGTM

Dmitriy-Litvinenko:
Hello @zburke @mkuklis Could you please review?

zburke (approved):
I am struggling a bit with the description here, but IIUC the problem is that we're showing an info toast instead of success test toast, which is inconsistent with how we handle "report is in progress" toasts in other apps. Is that correct? If so, this PR looks good, but please link to the UIIN story that describes the correct behavior.

Dmitriy-Litvinenko:
Hello @zburke  thank you for review. Yes you are right problem is that we're showing an info toast instead of success toast, which is inconsistent with other apps. Correct behavior was confirmed PO in https://issues.folio.org/browse/UIIN-1838 (more details in  
UIIN-1838 comment).

mkuklis (approved):

Dmitriy-Litvinenko:
Hello @zburke @mkuklis just for information we have failed github-actions-ci and how i understand we have similar problem for master branch.
image
I wrote about this in devops chanel.

PR #1513, opened 2021-11-11
UIIN-1654: Add permission for marking item as In process.

mkuklis:
Add permission for marking item as `In process`

https://issues.folio.org/browse/UIIN-1654


zburke (approved):
Yay for creating this pset _and_ pruning cruft at the same time!

PR #1518, opened 2021-11-22
UIIN-1854: After saving and updating a holdings record, the user is returned to the holdings instead of instance

BogdanDenis:
## Description
Fix issue when updating a Holdings record. After successful update user should see an updated record, but instead they see a Instance record first and Holdings record later.
This issue is probably caused by delays in `stripes-connect` between `UPDATE_SUCCESS` and `REFRESH` actions. Initially I tried adding a `isUpdating` property to `holdingsRecords` resource, but this delay made it difficult to write correct conditions to check when a record has been updated and refreshed.

## Approach
So eventually I decided to handle updating and refreshing Holdings record manually with a combination of `GET` and `reset` methods.
Added a `isLoadingUpdatedHoldingsRecord` which tells us if a record was updated and refreshed. In `isAwaitingResource` we're using this flag to show a loading message.
Also, calling `GET` in `didMount` to fetch Holdings record and call `reset` in `willUnmount` to clear `holdingsRecords.records` array otherwise when viewing several records user will only see the first loaded because of `holdingsRecords.records[0]`.

## Issues
[UIIN-1854](https://issues.folio.org/browse/UIIN-1854)

## Pre-Merge Checklist

### DIFF
--- a/src/ViewHoldingsRecord.js
+++ b/src/ViewHoldingsRecord.js
@@ -78,8 +78,10 @@ class ViewHoldingsRecord extends React.Component {
     holdingsRecords: {
       type: 'okapi',
       path: 'holdings-storage/holdings/:{holdingsrecordid}',
-      shouldRefresh: (resource, action, refresh) => {
-        return refresh || (action.meta.name === 'holdingsRecords' && action.meta.path === 'inventory');
+      resourceShouldRefresh: false,
+      accumulate: true,
+      GET: {
+        path: 'holdings-storage/holdings/:{holdingsrecordid}',
       },
       PUT: {
         path: 'inventory/holdings/:{holdingsrecordid}',
@@ -159,6 +161,8 @@ class ViewHoldingsRecord extends React.Component {
   }
 
   componentDidMount() {
+    this.props.mutator.holdingsRecords.GET();
+
     if (this.props.resources.instances1?.records[0]?.source) {
       if (this.isMARCSource() && !this.state.markRecord) {
         this.getMARCRecord();
@@ -176,6 +180,10 @@ class ViewHoldingsRecord extends React.Component {
         this.getMARCRecord();
       }
     }
+  }
+
+  componentWillUnmount() {
+    this.props.mutator.holdingsRecords.reset();
   }
 
   isMARCSource = () => {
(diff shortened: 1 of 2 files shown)


Vladyslav-Velytskyi (approved):

mariiaschv (approved):

src/ViewHoldingsRecord.js
```diff
@@ -220,6 +228,8 @@ class ViewHoldingsRecord extends React.Component {
     if (holdings.permanentLocationId === '') delete holdings.permanentLocationId;
     if (holdings.temporaryLocationId === '') delete holdings.temporaryLocationId;
 
+    this.setState({ isLoadingUpdatedHoldingsRecord: true });
```
mkuklis, line 231: @BogdanDenis would you mind explaining why we need this extra local state var?
The `accumulate` should still trigger `ACC_FETCH_SUCCESS` which flips `hasLoaded` and `isPending`. 

https://github.com/folio-org/stripes-connect/blob/9c53099c9bf09bf83723b988b0d07d8e52202fac/RESTResource/reducer.js#L41-L43

Could you use these flags instead?

BogdanDenis: @mkuklis Maybe I'm misunderstanding something, but `hasLoaded` is only `false` in initial state and is not reset after that. And I'm worried that changing it is probably going to break a lot of things in a lot of modules
But we still need this state variable because we need to know that Holdings record is loading because of an update or not. If we're loading after update - we need to wait for two requests to finish. Otherwise delay between PUT and GET requests will cause Instance record to appear

mkuklis: @BogdanDenis if it's not reset after that then it's a bug in stripes-connect we should address. You can also use `isPending` which seems to be working as expected.

BogdanDenis: @mkuklis `isPending` is working correctly when it's just one request, but when we have two requests (for update and for refresh) - there's a slight delay between them when `isPending` is false. So we need a flag that stays `true` throughout the whole time it takes to update and refresh the record.
Also, `isPending` is not enough to differentiate between loading a record when it's opened to view and loading it when it's been updated and needs to be refreshed.

mkuklis: @BogdanDenis sounds good. To me that local state is just a hack to go around the bug in stripes-connect. I understand that you want to move forward with this. Perhaps we can add a little comment to that code for now why this local state is required and go ahead and merge. I wonder if instead of using stripes-connect in the future we should try to use react-query.

mkuklis (approved):

PR #1523, opened 2021-12-06
UIIN-1620: Create title level request from Instance record

Stanislau-Snapok:
## Purpose
Add posibility to create title-level request from `Instance` record

## Approach
Done the same way as with `Item`: generates link to `request` module with `create` layer and pass `instanceId` in it.
New button appears only if TLR is switched on in settings, and we have permission to create requests.

## Refs
https://issues.folio.org/browse/UIIN-1620

## Screenshots
image

### DIFF
--- a/src/ViewInstance.js
+++ b/src/ViewInstance.js
@@ -1,6 +1,6 @@
 import { get } from 'lodash';
 import React, {
-  createRef
+  createRef,
 } from 'react';
 import PropTypes from 'prop-types';
 import { parse } from 'query-string';
@@ -34,8 +34,14 @@ import makeConnectedInstance from './ConnectedInstance';
 import withLocation from './withLocation';
 import InstancePlugin from './components/InstancePlugin';
 import { getPublishingInfo } from './Instance/InstanceDetails/utils';
-import { getDate, handleKeyCommand } from './utils';
-import { indentifierTypeNames, layers } from './constants';
+import {
+  getDate,
+  handleKeyCommand,
+} from './utils';
+import {
+  indentifierTypeNames,
+  layers,
+} from './constants';
 import { DataContext } from './contexts';
 
 import {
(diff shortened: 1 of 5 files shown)


pavelspichonak (approved):

Dmitriy-Litvinenko (approved):

mkuklis (approved):

src/ViewInstance.js
```diff
@@ -46,6 +52,15 @@ import {
 import { CalloutRenderer } from './components';
 
 import ImportRecordModal from './components/ImportRecordModal';
+import NewInstanceRequestBtn from './components/ViewInstance/MenuSection/NewInstanceRequestBtn';
```
zburke, line 55: This is super petty, but just call it `...Button` instead of `...Btn`.

src/components/ViewInstance/MenuSection/NewInstanceRequestBtn.js
```diff
@@ -0,0 +1,48 @@
...
+import { layers } from '../../../constants';
+
+export const getInstanceRequestLink = (instanceId) => `/requests?instanceId=${instanceId}&layer=${layers.CREATE}`;
+
+const NewInstanceRequestBtn = ({
+  isTlrEnabled,
+  instanceId,
+  intl: {
+    formatMessage,
+  },
```
zburke, line 20: You can also do `const intl = useIntl();` to get at the `intl` object via a hook instead of wrapping the exported component with `injectIntl`. You don't need to change this here, but I think it's considered the more idiomatic way to access it in a functional component.

zburke (approved):

src/ViewInstance.js
Stanislau-Snapok: Done.

src/components/ViewInstance/MenuSection/NewInstanceRequestBtn.js
Stanislau-Snapok: Good point, thanks! Done.

PR #1525, opened 2021-12-07
UIIN-1870 - Integrate facets with existing query input

mkuklis:
https://issues.folio.org/browse/UIIN-1870

This PR brings back facets and integrates them with the current query input. A side effect of this approach is that the `SearchAndSort` didn't have to be customized.

The PR also brings back the work done by @Baroquem around date filters in: https://github.com/folio-org/ui-inventory/pull/1494

inventory


Baroquem (approved):
Wow, nice work, Michal!

PR #1526, opened 2021-12-08
UIIN-1875 correctly specify eslint-related deps

zburke:
The `resolutions` entry `babel-eslint/@babel/parser` was a weird
typo/portmanteau of `@babel/eslint-parser` (the current thing) and
`babel-eslint` (the previous thing). Here, that bogus entry is being
removed, and `babel-eslint` is being replaced by `@babel/eslint-parser`.
There are some minor whitespace changes as well to accomodate the new
lint rules.

This PR _does not_ address the
rather-egregious-but-it-made-sense-at-the-time lint rule disabling
react-hooks checks ([UIIN-1415](https://issues.folio.org/browse/UIIN-1415)), but it does make dealing with it
possible. At the time, disabling it made sense because our old version
of lint totally barfed on some hooks-related syntax and so there was no
good way to keep that rule in place because it complained even when we
did the right thing with hooks' deps. Regrettably, it is now clear that
we are _not_ doing the right thing in many places. Alas. (CC:
@miketaylor).

Refs [UIIN-1875](https://issues.folio.org/browse/UIIN-1875)


mkuklis (approved):
👏

Baroquem (approved):

PR #1540, opened 2021-12-15
UIIN-1558: Fix bug preventing service point and source from appearing in item view circ history

Baroquem:
https://issues.folio.org/browse/UIIN-1558

The <ItemRoute> manifest uses a function to determine the right query to use when retrieving service point records, which are then used to populate details in the 'circulation history' section of <ItemView>. But if the service point to use wasn't yet available (i.e., the records were still loading), then the function returned `null`, causing the service point lookup to fail entirely. Returning an empty object instead of `null` fixes this.


mkuklis (approved):
Nice catch @Baroquem!

PR #1548, opened 2021-12-24
UIIN-1881: Inventory action menu link `View & reorder queue`

Stanislau-Snapok:
## Purpose
Add action menu link `View & reorder queue` for instance

## Approach
To show this button we muse enable TLR and have at least one request on this `instance`. To minimize the load we fetch only one request and use field `totalRecords`. Also in `request` module our route looks like `/requests/view/${requestId}/${instanceId}/reorder` and we have a lot of logic connected with `requestId` which we pass here.
I'am shure that pass of `requestId` is not the best way, but to make it another way we forced to do a lot of code refactoring.

## Refs
https://issues.folio.org/browse/UIIN-1881

## Screenshots
image

### DIFF
--- a/src/ViewInstance.js
+++ b/src/ViewInstance.js
@@ -41,6 +41,7 @@ import {
 import {
   indentifierTypeNames,
   layers,
+  REQUEST_OPEN_STATUSES,
 } from './constants';
 import { DataContext } from './contexts';
 
@@ -53,6 +54,7 @@ import { CalloutRenderer } from './components';
 
 import ImportRecordModal from './components/ImportRecordModal';
 import NewInstanceRequestButton from './components/ViewInstance/MenuSection/NewInstanceRequestButton';
+import RequestsReorderButton from './components/ViewInstance/MenuSection/RequestsReorderButton';
 
 const getTlrSettings = (settings) => {
   try {
@@ -61,6 +63,8 @@ const getTlrSettings = (settings) => {
     return {};
   }
 };
+const requestOpenStatuses = Object.values(REQUEST_OPEN_STATUSES);
+const instanceRequestsQuery = requestOpenStatuses.map(status => `status=="${status}"`).join(' OR ');
 
 class ViewInstance extends React.Component {
   static manifest = Object.freeze({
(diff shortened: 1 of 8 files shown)


src/ViewInstance.js
```diff
@@ -61,10 +63,20 @@ const getTlrSettings = (settings) => {
     return {};
   }
 };
+const requestOpenStatuses = Object.values(REQUEST_OPEN_STATUSES);
+const queryPartForInstanceRequests = requestOpenStatuses.map(status => `status=="${status}"`).join(' OR ');
```
pavelspichonak, line 67: better => instanceRequestsQuery

src/ViewInstance.js
```diff
@@ -466,22 +521,31 @@ class ViewInstance extends React.Component {
...
-                id="ui-inventory.viewRequests"
-                values={{ count: requests?.length ?? '' }}
-              />
-            </Icon>
-          </Button>
+          {
+            titleLevelRequestsFeatureEnabled
+              ? (
+                <RequestsReorderButton
+                  isUserHasPerms={canReorderInstanceRequests}
```
pavelspichonak, line 528: better => canReorder or hasReorderPermissions

pavelspichonak (approved):

src/ViewInstance.js
Stanislau-Snapok: Done, thanks!

src/ViewInstance.js
```diff
@@ -61,10 +63,20 @@ const getTlrSettings = (settings) => {
...
 };
+const requestOpenStatuses = Object.values(REQUEST_OPEN_STATUSES);
+const queryPartForInstanceRequests = requestOpenStatuses.map(status => `status=="${status}"`).join(' OR ');
 
 class ViewInstance extends React.Component {
   static manifest = Object.freeze({
     query: {},
+    configs: {
+      type: 'okapi',
+      records: 'configs',
```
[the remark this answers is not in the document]
Stanislau-Snapok: Done

src/ViewInstance.js
Stanislau-Snapok: Done

mkuklis (approved):

PR #1599, opened 2022-02-03
renaming

dmitriyfolio:
I've found this mismatching in process of smoke autotests writing. I suggest to fix with   spelling error.


BogdanDenis (approved):

mkuklis (approved):
Nice catch @dmitriyfolio!

PR #1606, opened 2022-02-10
UIIN-1831: Add pagination to holding items

mkuklis:
https://issues.folio.org/browse/UIIN-1831

This PR fetches all items attached to a given holdings record and adds in memory pagination to MCL used for these item records.

items


Baroquem (approved):
Nicely done!

PR #1615, opened 2022-02-24
UIIN-1645: Add 'all fields' search option

Baroquem:
https://issues.folio.org/browse/UIIN-1645

This implements the mod-search [search by all field values](https://github.com/folio-org/mod-search#search-by-all-field-values) option. Note that these options are disabled by default and have to be activated in the mod-search configuration, followed by a re-indexing.

Screen Shot 2022-02-24 at 4 07 35 PM


mkuklis (approved):
@Baroquem nicely done!

Do you think we could just remove `keyword all` option from the list?

https://github.com/folio-org/ui-inventory/pull/1615/files#diff-03423e533e7101fe103a71bd392486b92db5623a15d76ce0bbd3a96c39668017R109

I guest that would be a question for @CharlotteWhitt.

Baroquem:
> @Baroquem nicely done!
> 
> Do you think we could just remove `keyword all` option from the list?
> 
> I guest that would be a question for @CharlotteWhitt.

Yes, good question. I'll check with Charlotte.

Baroquem:
According to Charlotte, we're keeping both search options.

PR #1621, opened 2022-03-08
UIIN-1902: Fetch parent and child sub instances in one query

mkuklis:
https://issues.folio.org/browse/UIIN-1902

Previous implementation created a separate fetch request for every child and parent attached to the given instance object.
This was causing issues under chrome on window.

This PR changes the approach and creates a single request for all sub instances by using a query in a for of `(id==X OR id==Y ...)`


Baroquem (commented):
Hooray!

PR #1624, opened 2022-03-10
test some nodey things

zburke:
Output node version in GA workflow to help evaluate build errors. 

Errors in https://github.com/folio-org/ui-inventory/runs/5495824257?check_suite_focus=true imply GA is running NodeJS v17, which is causing a yarn v1 regression: https://github.com/nodejs/node/pull/42136#issuecomment-1063930789

### DIFF
--- a/.github/workflows/build-npm.yml
+++ b/.github/workflows/build-npm.yml
@@ -40,12 +40,19 @@ jobs:
 
       - name: Setup kernel for react native, increase watchers
         run: echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf && sudo sysctl -p
+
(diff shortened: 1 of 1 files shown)


zburke:
I'm so confused. Log output in this PR shows we are indeed on NodeJS 12, but the timing of the failed builds this morning, as mentioned in the description, lines up so nicely with [the bug reports in node](https://github.com/nodejs/node/issues/42279), and [the timing of the successful build later](https://github.com/folio-org/ui-inventory/actions/runs/1963116073) without any changes lines up so nicely with [the PR in node](https://github.com/nodejs/node/pull/42280), it feels like that must be implicated somehow.

PR #1630, opened 2022-03-21
Specify search.facets.collection.get UIIN-1941

adamdickmeiss:
Set module.inventory.enabled includes search.facets.collection.get
rather than search.instances.facets.collection.get.

This is part of interface search 0.7. That interface was not bumped
so it will be random whether ui-inventory will actually use the
right permission... Only can only rely on "pinned" interfaces in
platform-complete.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 # Change history for ui-inventory
 
 ## [9.1.0] IN PROGRESS
+* Use permission search.facets.collection.get rather than
+  search.instances.facets.collection.get. This was renamed some time
+  after interface search 0.7 was bumped. Breaking change that is
+  unfortunately not reflected in search interface. Fixes UIIN-1941.
+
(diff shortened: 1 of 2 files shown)


mkuklis (approved):

zburke (changes requested):
Since we know `search.facets.collection.get` wasn't introduced until `0.7`, please update the interface to `0.7` and add a "BREAKING change..." note to the CHANGELOG. 

IIUC, given when [these changes where made in mod-search](https://github.com/folio-org/mod-search/pull/165), `0.7` has been in place since before the version here was bumped to `v9.0.0`, so let's just leave the version as-is and acknowledge we forgot to add it to the `9.0.0` CHANGELOG at that time.

adamdickmeiss:
What needs to be done here? This is a real issue: "facets empty".

mkuklis:
@adamdickmeiss let's merge this in.

adamdickmeiss:
> @adamdickmeiss let's merge this in.

Done obviously.

PR #1633, opened 2022-03-23
UIIN-1966: Reset reference data resources on unmount.

mkuklis:
The resource data is being loaded every time the ui-inventory is opened. After leaving it and re-entering again the resources are re-fetched again but the previous data is still around (until new data arrives). This caused issues with stale data being served. 

This PR adds a `resourceShouldRefresh: true` which resets the resource data when the ui-inventory is unmounted so the fresh data is always being served.

https://issues.folio.org/browse/UIIN-1966

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,13 +2,12 @@
 
 ## [9.1.0] IN PROGRESS
 * The highlight of search results is not specific to the given search but now highlight all kinds of data in the record. Refs UIIN-1454.
-
 * Fetch parent and child sub instances in one query. Fixes UIIN-1902.
 * Missing Field - Receipt status under the Edit Holdings View. Fixes UIIN-1943.
-
 * updated "Date ordered" label to "Date opened". Refs UIIN-1946.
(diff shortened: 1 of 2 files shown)


marcjohnson-kint:
> After leaving it and re-entering again the resources are re-fetched again but the previous data is still around (until new data arrives).

What happens if the user does not leave and re-enter? For example, a different user creates a loan type in a separate browser on a different machine, will that get picked up without the first user re-entering the inventory UI?

mkuklis:
@marcjohnson-kint no the reference data is loaded only when the ui-inventory is loaded initially. Working from the different tab is not supported.

Baroquem (approved):

marcjohnson-kint:
> no the reference data is loaded only when the ui-inventory is loaded initially. Working from the different tab is not supported.

I think we need to make it explicit to the PO and other folks that even after this change, if a different person adds the loan type (or the same person from a different tab / window / browser) then it won't be picked up without a refresh.

If we don't explain that, I imagine folks will expect that this will work in all of those scenarios as well. At least, I would. I would think a primary use case is when a different member of staff does the admin then lets the cataloguing / circulation librarian know.

PR #1638, opened 2022-03-24
UIIN-1927: Fix confusing checkboxes in item move UI

Baroquem:
https://issues.folio.org/browse/UIIN-1927

This is not a good-looking UI, regardless of the approach used to fix the issue. These checkboxes are used to select which holdings records to move, so logically and aesthetically they would be better placed within the holdings accordion they indicate, rather than above it. However, placing them there triggers errors in `react-beautiful-dnd` and prevents drag and drop from working correctly. 

I found that the existing label for these checkboxes was (a) not being displayed, and (b) incorrectly describing selection of items rather than holdings. So I went for the most expedient solution of correcting the label and making sure that they appeared in the views. Still not a particularly good UI, but I hope it will be less confusing for the users.

Screen Shot 2022-03-24 at 5 44 29 PM

### DIFF
--- a/src/Instance/HoldingsList/Holding/Holding.js
+++ b/src/Instance/HoldingsList/Holding/Holding.js
@@ -23,20 +23,17 @@ const Holding = ({
   return (
     <div>
       {isDraggable &&
-        <FormattedMessage id="ui-inventory.moveItems.selectItem">
-          {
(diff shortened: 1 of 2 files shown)


mkuklis (approved):
@Baroquem I'm glad you were able to get this working. Nicely done!

PR #1648, opened 2022-04-04
UIIN-1997: Edit a MARC holdings record - generates a 500 error

mariiaschv:
## Purpose
Fix Edit a MARC holdings record generates a 500 error.

## Description
This bug occurred after editing MARC holdings record that was just created via quickmarc.
The problem was caused due to appending instance's `relatedRecordVersion` when redirecting to create holdings quickmarc page. Then when editing created MARC holdings record URL contained `relatedRecordVersion` of instance and then the `relatedRecordVersion` of the holdings record itself was appended.

## Issues
[UIIN-1997](https://issues.folio.org/browse/UIIN-1997)

### DIFF
--- a/src/ViewHoldingsRecord.js
+++ b/src/ViewHoldingsRecord.js
@@ -329,7 +329,10 @@ class ViewHoldingsRecord extends React.Component {
     const holdingsRecord = this.getMostRecentHolding();
 
     const searchParams = new URLSearchParams(location.search);
-    searchParams.append('relatedRecordVersion', holdingsRecord._version);
+
+    if (!searchParams.has('relatedRecordVersion')) {
+      searchParams.append('relatedRecordVersion', holdingsRecord._version);
+    }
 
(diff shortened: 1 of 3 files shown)


src/ViewHoldingsRecord.js
```diff
@@ -329,7 +329,10 @@ class ViewHoldingsRecord extends React.Component {
     const holdingsRecord = this.getMostRecentHolding();
 
     const searchParams = new URLSearchParams(location.search);
-    searchParams.append('relatedRecordVersion', holdingsRecord._version);
+
+    if (!searchParams.has('relatedRecordVersion')) {
```
BogdanDenis, line 333: Just as a precaution, to make sure that the parameter value is correct - let's delete the parameter and then add it. And same with `ViewInstance`

BogdanDenis (changes requested):

Svitlana-Korol (approved):

PR #1654, opened 2022-04-07
UIIN-1977: Reset search fields for each facet when Reset all button is clicked

mkuklis:
The state related to search values which belongs to each facet was previously
stored locally.

There was no easy way to reset it via `Reset all` because the `Reset all`
button and its action belongs to a completely different part of the DOM tree. 

This PR introduces [zustand](https://github.com/pmndrs/zustand) in order to
lift the local state up into a global store. The same could be accomplished via
react context or via local resource from stripes-connect but I find this
approach much simpler and cleaner.

https://issues.folio.org/browse/UIIN-1977

facets


Baroquem (approved):
This is an interesting approach. Zustand does look like a cleaner way of handling some of these state issues. Is it being used/considered in other parts of FOLIO?

PR #1655, opened 2022-04-08
UIIN-1988: Display correct locations in item form after edit

Baroquem:
https://issues.folio.org/browse/UIIN-1988

This component tries to maintain its own state to keep track of location changes in the item edit form, but it wasn't properly setting the initial value when the form first loaded (before any changes were made). If a user changed the location, saved and exited the form, then immediately re-opened the form, the old location would still be shown.

https://user-images.githubusercontent.com/1322242/162342148-809ee015-6a3f-49a0-a5e8-9b534bd1a18c.mov


mkuklis (approved):

zburke (commented):
@Baroquem, I'm glad this works, but given that `selectedValue` is initialized to `value` in LocationSelectionWithCheck:
```
const [selectedValue, setSelectedValue] = useState(value);
```
I don't understand why this changes things. I feel like I'm missing something fundamental about lifecycle methods here :/

Baroquem:
@zburke I think the issue is that the state is initialized the first time the component is mounted, but then it isn't changed again unless the user manually selects a different location in the dropdown list. Thus there's no mechanism to set the state correctly when the form is first (re)loaded, before a selection is made.

PR #1658, opened 2022-04-13
UIIN-1980: Update locations in <ViewHoldingsRecord> after edit.

Baroquem:
https://issues.folio.org/browse/UIIN-1980

This commit includes a lot of cleanup of location-related code that was using local resources to manage location state. Because the resources weren't being updated properly, the locations displayed in the location accordion would never be updated after an edit unless the entire page were refreshed in the browser. I discovered that the warning modals shown before deletion, which also state the holdings location, were likewise using stale location data.

The pane header, on the other hand, displayed the correct location because it was using passed-in referenceTables to determine location name from the holdings record. Switching to that method for the location accordion as well allows us to eliminate a lot of unnessary code.

I don't _think_ that this refactoring should have any bad side effects!


mkuklis (approved):
❤️ 
Every time I see `getDerivedStateFromProps` being removed it makes me happy :)

Thank you for cleaning this up @Baroquem! Great work.

src/ViewHoldingsRecord.js
```diff
@@ -492,8 +455,8 @@ class ViewHoldingsRecord extends React.Component {
     const instanceSource = referenceTables?.holdingsSources?.find(source => source.name === instance.source);
     const holdingsRecord = this.getMostRecentHolding();
     const holdingsSource = referenceTables?.holdingsSources?.find(source => source.id === holdingsRecord.sourceId);
-    const holdingsPermanentLocation = holdingsRecord.permanentLocationId ? permanentLocation.records[0] : null;
-    const holdingsTemporaryLocation = holdingsRecord.temporaryLocationId ? temporaryLocation.records[0] : null;
+    const holdingsPermanentLocation = get(referenceTables?.locationsById[holdingsRecord?.permanentLocationId], ['name'], '-');
+    const holdingsTemporaryLocation = get(referenceTables?.locationsById[holdingsRecord?.temporaryLocationId], ['name'], '-');
```
zburke, line 459: `<NoValue>` > `'-'`

zburke (approved):
Oh man, this is so much cleaner, and the cruft you pruned is "initial commit" code from 2017! Thank you!

PR #1669, opened 2022-05-06
UIIN-2030: Inventory app white screen when no tags exist

mariiaschv:
## Purpose
Fix inventory app white screen when no tags exist.

This problem was caused by checking resource's `hasLoaded` property. In the case of failed request `hasLoaded` stayed `false` leading to the infinite loading. The solution was to check `hasLoaded`, `isPending`, and `failed` properties together to determine requests that are just loading but did not failed.

## Screenshots
chrome_sa9WdWU2wb

## Issues
[UIIN-2030](https://issues.folio.org/browse/UIIN-2030)


BogdanDenis (approved):

Svitlana-Korol (approved):

mkuklis (approved):
@mariiaschv nicely done 👏 !

PR #1673, opened 2022-05-16
UIIN-2046: Inventory app crashes when locations are not loaded

BogdanDenis:
## Description
When location request either fails for some reason, or has empty response and a Holdings record tries to find location name by it's `permanentLocationId` - the app crashes

## Issues
[UIIN-2046](https://issues.folio.org/browse/UIIN-2046)

PR #1674, opened 2022-05-16
UIIN-2030: Adjust isPending check for reference data resources

mkuklis:
The issue with the previous attempt was that during the first render the resources are still in the initial state:

````js
hasLoaded: false
failed: false
isPending: false
````

causing the `isLoading` to return `false`.

This PR changes it a bit and it will return `isPending === true` if the resource is marked as `isPending` or if it did not start the fetching yet (it is still in the initial state). 

[UIIN-2030](https://issues.folio.org/browse/UIIN-2030)

### DIFF
--- a/src/providers/DataProvider.js
+++ b/src/providers/DataProvider.js
@@ -18,11 +18,13 @@ const DataProvider = ({
   const isLoading = useMemo(() => {
     // eslint-disable-next-line guard-for-in
     for (const key in manifest) {
-      const isRecourceLoading = !(resources?.[key]?.hasLoaded)
-        && resources?.[key]?.isPending
-        && !(resources?.[key]?.failed);
+      const resource = resources?.[key] ?? {};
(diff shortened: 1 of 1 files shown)


zburke (approved):

BogdanDenis:
Hi @mkuklis, for some reason this causes infinite re-requesting resources when opening an Instance record.
There's also lots of mounting and unmounting in `connect`, and I'm not very familiar with that area so I don't understand why it's happening.
This is only happening when using `isPending` in checking, and the condition makes total sense and it looks like the flags have correct values
I'd fixed it in [my PR](https://github.com/folio-org/ui-inventory/pull/1673), but I'm not using `isPending` there

mkuklis:
Thanks @BogdanDenis. Yes something is still off here. Things worked just fine for me locally. Let me pull your PR here and test it a bit. If it works we should merge it.

mkuklis:
Oh it looks like it was merged already. That's good since the other approach was still broken. I will pull it and try it.

PR #1684, opened 2022-05-19
lock @vue/compiler-sfc

zburke:
Spike: lock `@vue/compiler-sfc` to `3.2.33` to avoid the buggy 3.2.34 release to see if vuejs/core/issues/5957 is the source of our build failures on our CI machines that are still running NodeJS v12.

### DIFF
--- a/package.json
+++ b/package.json
@@ -909,6 +909,7 @@
     "@folio/plugin-find-instance": "^6.0.0"
   },
   "resolutions": {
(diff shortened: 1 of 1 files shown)


zburke:
Nope; gonna upgrade NodeJS to v16 instead, see [UIIN-2049](https://issues.folio.org/browse/UIIN-2049).

PR #1686, opened 2022-05-20
provide CalloutContext.Provider in the test harness

zburke:
Wrap the component in `<CalloutContext.Provider>` with a dummy
`sendCallout` prop so that tests that render components that leverage
context don't blow up with NPEs. Locally, we'd see warnings in the test
output like
```
[UnhandledPromiseRejection: This error originated either by throwing
inside of an async function without a catch block, or by rejecting a
promise which was not handled with .catch(). The promise rejected
with the reason “TypeError: Cannot read properties of undefined
(reading ‘sendCallout’)“.] {
  code: ‘ERR_UNHANDLED_REJECTION’
}
```
but tests would pass; in CI, however, such errors would cause the test
suite to abort.

This was discovered while working on UIIN-2049 to bump the CI Node
version to v16.


mkuklis (approved):
❤️

PR #1708, opened 2022-06-03
Revert "UIIN-2045 Fix browse form refresh issue (#1700)"

zburke:
This reverts commit da5f59eae201ff15a7af899cab4681489d567613.

Unfortunately, #1700 caused as many problems as it solved. With those
changes in place, regular searches (not browses) that resulted in a
single record would cause the results list and details pane to be stuck
in an infinite refresh loop.

Refs [UIIN-2045](https://issues.folio.org/browse/UIIN-2045)


uladislausamets (approved):
I'm okay with reverting these changes, but we need some workaround how to fix the issue described in UIIN-2045

Baroquem (approved):

PR #1712, opened 2022-06-11
UIIN-2062: Name Type facet sends requests to incorrect endpoint

mariiaschv:
## Purpose
Fix facets send requests to incorrect endpoints.

Related changes:
- get rid of `browse` segment and make `browse` a part of `instance` segment.
- render correct filters in `instanceFilterRenderer.js` depending on which option was selected.
- the anchor row doesn't show up if a facet has been applied but no search param provided.

## Screenshots
https://user-images.githubusercontent.com/84023879/173166801-09c0ba9e-5d5a-4b5b-8cba-735601354a28.mp4

image

## Issues
[UIIN-2062](https://issues.folio.org/browse/UIIN-2062)


mkuklis (commented):
@mariiaschv thank you for your take on this. The PR includes some good changes and fixes in `useFacets` and `MultiSelectionFacet`.

However there is something fundamentally wrong with this approach and I think it actually started in other PRs related to `browse`.

The issue here is that a new segment called `browse` was introduced previously which is a wrong abstraction for this setup.

The original `segment` suppose to represent one of the tabs `instance`, `holdings` and `item`. The `browse` should actually sit under the `instance` segment and it should not be represented in the code as a segment at all.

Perhaps it should be renamed to `searchMode` or `querySearchMode`.

We really should not introduce a new segment here:

https://github.com/folio-org/ui-inventory/blob/master/src/filterConfig.js#L311

and rather try to handle what filters should be rendered in the lower level (in `instanceFilterRenderer`) based on the current `searchMode` (or `qindex`).

I realized that this is not related to your work here but it feels like you are building on top of already wrong abstraction. I think most of this PR looks fine. My only issue is really related to 
what happens in:

`InstanceList.componentDidUpdate`

Which I think touch on what I described above.

mariiaschv:
> The original `segment` suppose to represent one of the tabs `instance`, `holdings`, and `item`. The `browse` should actually sit under the `instance` segment and it should not be represented in the code as a segment at all.

@mkuklis thank you for the thorough comment and your explanations! I'll try to rework it.

mariiaschv:
> Regarding `InstanceList.componentDidUpdate`, it handles the case when the facet is applied without the search query. In this case we don't need to show the anchor row, and thus we should filter it out.

upd: I removed `InstanceList.componentDidUpdate` and used `highlightMatch` parameter instead.

mkuklis (approved):
@mariiaschv thank you for taking extra time with this. It looks great! 🎉

Svitlana-Korol (approved):

uladislausamets (approved):

BogdanDenis (approved):

PR #1736, opened 2022-07-12
UIIN-2087 Fix Non-exact match placeholder message displayed when user switching between browse contributors result list pages.

BogdanDenis:
## Description
When user clicks on "Prev" or "Next" button - we're updating `query` parameter in the url. That parameter was used to display empty result message
But the url is updated instantly after clicking, so it looks like new query is not found.
Fixed the issue by displaying search term in response.

## Screenshots

https://user-images.githubusercontent.com/19309423/178442259-df3f10a8-83da-4aeb-91ed-69ca80907c70.mp4

## Issues
[UIIN-2087](https://issues.folio.org/browse/UIIN-2087)


Dmytro-Melnyshyn:
Is it ok for the search field to change the value after clicking on the `Previous` button?

BogdanDenis:
> Is it ok for the search field to change the value after clicking on the `Previous` button?

Input value should change, and https://github.com/folio-org/ui-inventory/pull/1738 will remove parameter name from input

ODanyor (approved):

UladzislauKutarkin (approved):

mkuklis (approved):

PR #1737, opened 2022-07-12
UIIN-2094: browse contributors may show results floating amid blank rows

Dmytro-Melnyshyn:
## Purpose
- changing a select-option should not submit a form
- results pane shouldn't show blank rows

## Approach
- as for changing a "select-option should not submit a form": use the 'fetch' function to determine if the fetch should execute or not.
- as for empty rows: the column headings are different for the browse options, so the rows are displayed without data. Now after selecting any browse option, the records will be reset.

https://user-images.githubusercontent.com/77053927/179087298-263f85ea-7fba-4dc7-a88f-0ae9a7f6e09f.mp4

## Pre-Merge Checklist


BogdanDenis (commented):
I think adding a new resource to manifest and related logic might be an overkill - it seems that the problem is that url is updated when user in making changes to index, even though it shouldn't happen.
We can fix it be removing `qindex` from `parentMutator.query.update` call inside `onChangeIndex`
This way local state will be updated and url won't change until submit button is clicked
Could you check if this solution works please?

Dmytro-Melnyshyn:
> I think adding a new resource to manifest and related logic might be an overkill - it seems that the problem is that url is updated when user in making changes to index, even though it shouldn't happen. We can fix it be removing `qindex` from `parentMutator.query.update` call inside `onChangeIndex` This way local state will be updated and url won't change until submit button is clicked Could you check if this solution works please?

Thanks for the comment. It will work if filters are not selected. `parentMutator.query.update` also removes filters from the URL.

BogdanDenis (approved):

mkuklis:
@Dmytro-Melnyshyn the `fetch` could be also a function which can be used to determine if the fetch should execute or not (by returning `true` or `false`). It receives current props of the connected component. Perhaps there is something in the props which can be used to control the fetch? Have you explored that approach? It would be really nice not to introduce that extra local resource (`manifestFetchPropToSearch`).

mkuklis (approved):

Dmytro-Melnyshyn:
> @Dmytro-Melnyshyn the `fetch` could be also a function which can be used to determine if the fetch should execute or not (by returning `true` or `false`). It receives current props of the connected component. Perhaps there is something in the props which can be used to control the fetch? Have you explored that approach? It would be really nice not to introduce that extra local resource (`manifestFetchPropToSearch`).

@mkuklis thanks for the comment, I've changed the approach, please have a look.

BogdanDenis (approved):

mkuklis (approved):
@Dmytro-Melnyshyn thank you for exploring this further. Nice work!

PR #1748, opened 2022-07-21
UIIN-2111: Fix import paths for stripes.

mkuklis:
https://issues.folio.org/browse/UIIN-2111

Use correct import paths for stripes modules.


zburke (changes requested):
Weird, I'm not sure how the import change is affecting these two tests, but it is: the failures reproduce locally for me 😭 . Ofc, the import changes themselves look fine. LMK if you want help investigating.

mkuklis:
@zburke I fixed that one failing test. It looks like some test files are still using `stripes-core` reference but I think we can clean them up in a separate PR.

zburke (approved):
It's still weird to me that switching the import changes the test behavior. I guess, somehow, the stripes-core mock of `useOkapiKy` is now insufficient??? Weird. Regardless, I'm delighted (and impressed!) that you figured it out!

mkuklis:
@zburke `useOkapiKy` was already mocked below with:

````js
    useOkapiKy
      .mockClear()
      .mockReturnValue(kyMock);
````
(It's hard to see it in the PR because it didn't change)

Because of that we didn't really need `jest.mock('@folio/stripes-core/src/useOkapiKy', () => jest.fn());`

It looks like `jest` is treating these two:

`import useOkapiKy from '@folio/stripes-core/src/useOkapiKy';`

and

`import { useOkapiKy } from '@folio/stripes/core';`

differently and since the implementation is now using `import { useOkapiKy } from '@folio/stripes/core';`

I think jest was mocking a wrong `useOkapiKy`. Still strange. I agree.

zburke:
> since the implementation is now using `import { useOkapiKy } from '@folio/stripes/core';` I think jest was mocking a wrong `useOkapiKy`

In fact, that makes perfect sense. Thanks for the explanation, @mkuklis! It's great to see this work wrapping up so we can move along with migrating to npm!

PR #1749, opened 2022-07-25
UIIN-2124: The "Instance" record still displayed at the third pane when user selects "Browse contributors" option

ODanyor:
## Purpose
The "Instance" record still displayed at the third pane when user selects "Browse contributors" option

## Approach
Redirect user to `inventory/` on `qindex` change to any browse options to close record view window.

## Issues
[UIIN-2124](https://issues.folio.org/browse/UIIN-2124)

### DIFF
--- a/src/components/InstancesList/InstancesList.js
+++ b/src/components/InstancesList/InstancesList.js
@@ -781,6 +781,7 @@ class InstancesList extends React.Component {
       match: {
         path,
       },
+      goTo,
       namespace,
       stripes,
       fetchFacets,
@@ -967,6 +968,7 @@ class InstancesList extends React.Component {
       if (isBrowseOption) {
         parentMutator.browseModeRecords.reset();
         this.setState({ isSingleResult: false });
+        goTo(path);
       } else {
         this.setState({ isSingleResult: true });
       }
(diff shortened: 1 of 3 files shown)


src/components/InstancesList/InstancesList.js
```diff
@@ -967,6 +968,7 @@ class InstancesList extends React.Component {
       if (isBrowseOption) {
         parentMutator.browseModeRecords.reset();
         this.setState({ isSingleResult: false });
+        goTo(path);
```
BogdanDenis, line 971: Will this also close the record when switching from one browse option to another browse option?

ODanyor: I was worrying  about that too. But the list items don't open any record, they redirect user to non browse options. Correct me If I am wrong

BogdanDenis (approved):

src/components/InstancesList/InstancesList.js
Dmytro-Melnyshyn, line 971: I think it is necessary to leave the `qindex` param in the URL.

ODanyor: Yes, you are right. Thank you for noticing that!

BogdanDenis (approved):

PR #1751, opened 2022-07-26
UIIN-2130: valid prop type for regExpForQuery

ODanyor:
## Purpose
Failed prop type: Invalid prop `regExpForQuery` of type `regexp` supplied to `SearchAndSort`, expected `string`.


BogdanDenis:
Converting regex to string will break the functionality, we should change the prop type instead.
I'll create a PR in SSC instead

PR #1758, opened 2022-08-05
UIIN-2125: The browse query changed when user returns from search to browse contributors pane

Dmytro-Melnyshyn:
## Purpose
1. The search field value should have the search term entered by the user after opening a new browser tab and inserting the search bar.
2. The value of the search field should remain the same after doing a `browse` search and clicking on the search result, and then going back by clicking the browser back arrow.

## Approach
~~When the user clicks on the prev/next buttons below the list of results, the `query` loses the search term entered by the user, so the search field value will not be correct after opening a new browser tab and pasting the search bar. Therefore, a new parameter has been added to the URL, which will store the user's search query and appear only by clicking on the prev/next buttons.~~

~~To achieve the second purpose, the `syncQueryWithUrl` prop is passed to the `SearchAndSort` component. And now the value of `locallyChangedSearchTerm` is taken from the new parameter `userQuery` (if `query` doesn't match the search query entered by the user) or, if the `query` matches, then from `query`.~~

When the user clicks on the prev/next buttons below the list of results, the `query` loses the search term entered by the user, so the search field value will not be correct after opening a new browser tab and pasting the search bar. Therefore, a new `browsePoint` parameter has been added to the URL, which will hold the `browse` offset since the `query` is already taken up by the searched-for value.

`Browse` queries look like

```
qindex=contributor&query=snow
qindex=contributor&query=snow&browsePoint=name%20<%20"Odey%2C%20Mike%20Odugbo"
qindex=contributor&query=snow&browsePoint=name%20>%20"Novik%2C%20Naomi"
```

where a user has clicked the `Browse` button on the "Search & filter" pane, the `< Previous` link, and the `Next > ` link, respectively.

Also, changes have been added to the `stripes-smart-components` [STSMACOM-697](https://github.com/folio-org/stripes-smart-components/pull/1276)

## Issues
[UIIN-2125](https://issues.folio.org/browse/UIIN-2125)

## Screencast

https://user-images.githubusercontent.com/77053927/183582616-d325a3ef-fd24-4728-8561-ffaacd55b81f.mp4

## Pre-Merge Checklist


zburke (commented):
I feel like this PR kind of gets things backwards. The explanation you've given about why the `query` query parameter loses its value when navigating through "browse" results makes sense, but rather than concluding "so we need `userQuery` to hold the original value" as you do here, I conclude "we need `somethingLikeBrowsePoint` to hold the browse offset since `query` is already taken up by the searched-for value".

Maybe browse queries could look like
```
qindex=contributor&query=snow&view=browse
qindex=contributor&query=snow&view=browse&browsePoint=name%20<%20"Odey%2C%20Mike%20Odugbo"
qindex=contributor&query=snow&view=browse&browsePoint=name%20>%20"Novik%2C%20Naomi"
```
where a user has clicked the "Browse" button on the (now misnamed) "Search & filter" pane, the "< Previous" link, and the "Next > " link, respectively. 

The bottom line is, every other app in the entire system uses `query` to hold "the thing the use searched for" and I think that's a convention worth following. If that's no longer practical here because too much work has already been done, that's unfortunate, and I won't stand in the way of merging this. But I don't really like it.

Dmytro-Melnyshyn:
> I feel like this PR kind of gets things backwards. The explanation you've given about why the `query` query parameter loses its value when navigating through "browse" results makes sense, but rather than concluding "so we need `userQuery` to hold the original value" as you do here, I conclude "we need `somethingLikeBrowsePoint` to hold the browse offset since `query` is already taken up by the searched-for value".
> 
> Maybe browse queries could look like
> 
> ```
> qindex=contributor&query=snow&view=browse
> qindex=contributor&query=snow&view=browse&browsePoint=name%20<%20"Odey%2C%20Mike%20Odugbo"
> qindex=contributor&query=snow&view=browse&browsePoint=name%20>%20"Novik%2C%20Naomi"
> ```
> 
> where a user has clicked the "Browse" button on the (now misnamed) "Search & filter" pane, the "< Previous" link, and the "Next > " link, respectively.
> 
> The bottom line is, every other app in the entire system uses `query` to hold "the thing the use searched for" and I think that's a convention worth following. If that's no longer practical here because too much work has already been done, that's unfortunate, and I won't stand in the way of merging this. But I don't really like it.

Great idea, thanks @zburke! Corrected.

Dmytro-Melnyshyn:
@mkuklis, @BogdanDenis, @UladzislauKutarkin please take a look.

BogdanDenis (approved):
Looks good! Just one comment in the related SSC PR

uladislausamets (approved):

BogdanDenis (approved):

mkuklis (approved):

mkuklis:
@Dmytro-Melnyshyn thank you for addressing the concerns from @zburke. Nice work!

PR #1783, opened 2022-09-14
UIIN-2189 - Reset qindex inside buildQuery

mkuklis:
Reset `qindex` when a query is being built for a given facet. This is required otherwise `makeQueryFunction` won't use `queryTemplate` and will fall back to a provided query value.

https://github.com/folio-org/stripes-smart-components/blob/e918a620ad2ac2c5b06ce121cd0e061a03bcfdf6/lib/SearchAndSort/makeQueryFunction.js#L46

Thank you @zburke for catching this issue.

https://issues.folio.org/browse/UIIN-2189

### DIFF
--- a/src/components/InstancesList/InstancesList.test.js
+++ b/src/components/InstancesList/InstancesList.test.js
@@ -3,7 +3,12 @@ import { Router } from 'react-router-dom';
 import { noop } from 'lodash';
 import userEvent from '@testing-library/user-event';
 import { createMemoryHistory } from 'history';
-import { screen, fireEvent, cleanup } from '@testing-library/react';
+import {
+  cleanup,
+  fireEvent,
+  screen,
+  waitFor,
+} from '@testing-library/react';
 
 import '../../../test/jest/__mock__';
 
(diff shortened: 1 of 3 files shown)


zburke (approved):
LGTM by whyohwhy don't the tests pass in CI? They pass locally!

src/components/InstancesList/InstancesList.test.js
```diff
@@ -141,27 +143,18 @@ describe('InstancesList', () => {
       expect(document.querySelectorAll('#pane-results-content .mclRowContainer > [role=row]').length).toEqual(3);
     });
 
-    it('should have selected browse call number option', () => {
-      fireEvent.change(screen.getByRole('combobox'), {
-        target: { value: 'callNumbers' }
-      });
-
+    it('should have selected browse call number option', async () => {
+      await userEvent.selectOptions(screen.getByLabelText('Search field index'), 'callNumbers');
```
JohnC-80, line 147: looks like switching to `userEvent` might have caused things to progress faster than they previously were.
I can see these failures locally, and wrapping the `expect` statement in a `waitFor` call allowed them enough time to pass.
```
waitFor(() => { 
  expect((screen.getByRole('option', { name: 'Browse call numbers' })).selected).toBeTruthy(); 
});
```

JohnC-80: I'm uncertain, but the timing difference may also have come from the lifecycle refactor in searchAndSort - though I'm not sure on the specifics - the select changes the state which feeds it back is value... this is re-confirmed when the query parameters come back to the component, so... 🤷

zburke: That's interesting info, @JohnC-80, thanks. I've also found that I can repeat the failures locally using the `yarn.lock` from CI. I'm trying to figure out if I can narrow the problem to a specific version change now.

JohnC-80 (approved):

src/components/InstancesList/InstancesList.test.js
zburke: `waitFor` FTW. Thanks, @JohnC-80! Still, I want to figure out why this broke. Building with an old `yarn.lock` file broke tests, but then pinning to a newer stripes-smart-components left it still broken, suggesting something else is the culprit. 

Regardless, @mkuklis, this is good to merge now!

mkuklis:
@zburke thank you!

PR #1785, opened 2022-09-16
UIIN-2191 No results return when you conduct a Contributor search

BogdanDenis:
## Description
Use `==/string` search when clicking on Browse contributors result, but just `=` search when searching contributors

## Additional changes
Downgraded `history@5` to `history@4` due to incompatibility with `react-router@5`
Application behaves correctly because `<Root>` inside `stripes-core` uses `history@4`
But module level tests fail because `ui-inventory` uses `history@5`.

The source incompatibility lies in how `history` notifies `react-router` that location has changed.
Here's how `react-router` listens to this change: https://github.com/remix-run/react-router/blob/v5.3.3/packages/react-router/modules/Router.js#L47.
It accepts a single `location` argument that is saved inside state as `location`

`history@4` notifies listeners like so:
https://github.com/remix-run/history/blob/v4.10.1/modules/createBrowserHistory.js#L85
which calls https://github.com/remix-run/history/blob/3f69f9e07b0a739419704cffc3b3563133281548/modules/createTransitionManager.js#L66
It calls the handler and destructures the arguments array

But this is how `history@5` notifies listeners:
https://github.com/remix-run/history/blob/v5.3.0/packages/history/index.ts#L481

It calls the handler with object `{ location, action }`

So when `react-router` handles the call, `location` is not 
```
{
  pathname: ...,
  search: ...,
  ...etc
}
```
but instead
```
{
  action: 'PUSH', // or POP or whatever action
  location: {
    pathname: ...,
    search: ...,
    ...etc
  },
}
```

Hey @zburke, this is probably fixed in `react-router@6` because it accepts `history@5` in it's dependencies
https://github.com/remix-run/react-router/blob/v6.3.0/package.json#L44
Are we planning to upgrade to `react-router@6` any time soon?
Also, we definitely should not update `history` in `stripes-core`

## Screenshots

https://user-images.githubusercontent.com/19309423/190616364-6aef7b17-9705-4067-88f7-09a8e2f4f1b9.mp4

## Issues
[UIIN-2191](https://issues.folio.org/browse/UIIN-2191)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@
 * Escape quotes in search string. Fix UIIN-2143.
 * Introduce in-memory pagination when loading parent/child instances. Fixes UIIN-2155.
 * The placeholder for missing match is clickable on the Browse list. Fixes UIIN-2126.
+* No results return when you conduct a Contributor search. Fixes UIIN-2191.
 
 ## [9.1.0](https://github.com/folio-org/ui-inventory/tree/v9.1.0) (2022-06-28)
 [Full Changelog](https://github.com/folio-org/ui-inventory/compare/v9.0.0...v9.1.0)
--- a/package.json
+++ b/package.json
@@ -886,7 +886,7 @@
     "file-saver": "^2.0.0",
     "final-form": "^4.18.2",
     "final-form-arrays": "^3.0.1",
-    "history": "^5.0.0",
+    "history": "^4.10.0",
     "ky": "^0.23.0",
     "lodash": "^4.17.4",
     "moment": "~2.24.0",
--- a/src/constants.js
+++ b/src/constants.js
@@ -140,7 +140,8 @@ export const holdingsStatementTypes = [
 export const queryIndexes = {
   SUBJECT: 'subject',
   QUERY_SEARCH: 'querySearch',
-  CALL_NUMBER: 'callNumber'
+  CALL_NUMBER: 'callNumber',
+  CONTRIBUTOR: 'contributor',
 };
 
 export const indentifierTypeNames = {
--- a/src/filterConfig.js
+++ b/src/filterConfig.js
@@ -115,7 +115,7 @@ export const instanceIndexes = [
   // instead. It might make sense to rename the keyword option to something like `keywordAll`
   // but, without tracing the use of the value, I don't know what effects that would have in the code.
   { label: 'ui-inventory.search.all', value: 'all', queryTemplate: 'keyword all "%{query.query}" or isbn="%{query.query}"' },
-  { label: 'ui-inventory.contributor', value: 'contributor', queryTemplate: 'contributors.name=="%{query.query}"' },
+  { label: 'ui-inventory.contributor', value: 'contributor', queryTemplate: 'contributors.name="%{query.query}"' },
   { label: 'ui-inventory.title', value: 'title', queryTemplate: 'title all "%{query.query}"' },
   { label: 'ui-inventory.identifierAll', value: 'identifier', queryTemplate: 'identifiers.value="%{query.query}"  or isbn="%{query.query}"' },
   { label: 'ui-inventory.isbn', value: 'isbn', queryTemplate: 'isbn="%{query.query}"' },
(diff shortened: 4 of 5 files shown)


Dmytro-Melnyshyn (approved):

odoniyor (approved):

zburke:
@BogdanDenis, wow, impressive sleuthing on the `history` changes between v4 and v5. AFAIK, we haven't done any actual testing with react-router v6, but my general understanding is that the v5 to v6 change is a big one. We have discussed many major-version upgrades for peer-deps at recent #stripes-architecture meetings, and given that the list is long and (until now) we haven't seen many problems with react-router we have generally pushed it to the bottom of the list. 

All that said, it feels to me like the "real" outcome of your discovery here is `history` should be a peer-dep in apps and a direct-dep at the platform level, with the corollary that we need to agree on a single version (presumably v5) since apps and `stripes-core` both need to agree on the API. Of course, adding a peer-dep is a breaking change. Ooof. 

Looking through platform-complete shows that in fact `react-router` v5 is the most prevalent:
```
$ grep -r '"history"' ./node_modules/@folio/*/package.json | sed -E 's/^(.*):(.*):(.*)$/\3:\2:\1/' | sort
 "^4.6.3",:    "history":./node_modules/@folio/plugin-eusage-reports/package.json
 "^4.6.3",:    "history":./node_modules/@folio/stripes-core/package.json
 "^5.0.0",:    "history":./node_modules/@folio/data-import/package.json
 "^5.0.0",:    "history":./node_modules/@folio/eholdings/package.json
 "^5.0.0",:    "history":./node_modules/@folio/finance/package.json
 "^5.0.0",:    "history":./node_modules/@folio/inn-reach/package.json
 "^5.0.0",:    "history":./node_modules/@folio/inventory/package.json
 "^5.0.0",:    "history":./node_modules/@folio/marc-authorities/package.json
 "^5.0.0",:    "history":./node_modules/@folio/quick-marc/package.json
 "^5.0.0",:    "history":./node_modules/@folio/stripes-authority-components/package.json
 "^5.0.0",:    "history":./node_modules/@folio/users/package.json
 "^5.0.1",:    "history":./node_modules/@folio/oai-pmh/package.json
 "^5.1.0",:    "history":./node_modules/@folio/bulk-edit/package.json
 "^5.1.0",:    "history":./node_modules/@folio/tenant-settings/package.json
```
which makes it tempting in the short-term to apply changes in `stripes-core` (and `ui-plugin-eusage-reports`) rather than here. On the one hand, changing `stripes-core` potentially has a wider impact since everything depends on it, but maybe it's the better change since it would bring versions into alignment across more applications, i.e. it feels "more correct". 

What do you think? I'm open to merging this PR as-is, but also think it's worth zooming out and considering other approaches.

UladzislauKutarkin (approved):

BogdanDenis:
@zburke 

> which makes it tempting in the short-term to apply changes in stripes-core (and ui-plugin-eusage-reports) rather than here

Now I realize I said "we definitely should not update history in stripes-core" in my original comment, but should have said "should hot **upgrade**", because upgrading from v4 to v5 introduces this issue

So the "good" version of history is v4

It seems that we have several options:
1. Upgrade `react-router` to v6
A lot of work and we're not ready to do this at this point
2. Make `history` a peer-dep.
I assume it is still a breaking change to introduce a peer-dep even when this peer-dep will have version `^4.10.0`, same as current version in `dependencies`, right?
3. Upgrade history to v5 in `stripes-core` and `ui-plugin-eusage-reports`.
This will align the versions in all modules, but seeing as `history@5` is incompatible with `react-router@5` we shouldn't do this
4. Downgrade `history` to v4 in all modules.
A lot of tickets and PRs for different teams

I think short-term we should go with option 4, which will other teams from getting this issue in tests that I faced
Then mid-term perhaps option 2, so all modules use the same version of `history`
And long-term - option 1. It will let us use `react-router@6` features and unfreeze `history`

zburke:
>  seeing as history@5 is incompatible with react-router@5

Ahh, I didn't understand this at first. Somehow, instead of reading it as "these two versions of _third-party_ libraries are incompatible" I thought "we just need to change _our code_ to accommodate the v4 to v5 API changes in `history`". Thank you for your patience and for explaining it, again. 

Are you able to attend #stripes-architecture on Thursday at 10 ET/14 UTC to discuss the options? It's true that we'd been planning to delay the react-router v6 upgrade but AFAIK we haven't actually done a POC to see what will be involved, we just thought, "ooooh, looks big, let's wait because v5 is working fine for us." But my understanding from this PR is that due to the history/react-router version incompatibility we have introduced, it's likely that v5 may not be working as well as we think but we just don't know it 😬 . Given that, maybe it makes more sense to investigate react-router v6 rather than going through two rounds of changes (downgrading all repos to history to v4, then upgrading all repos to history v5 and react-router v6).

BogdanDenis:
Sure, @zburke, I'll join the meeting

PR #1805, opened 2022-10-03
UIIN-1968: Holdings: show inactive by location name

maccabeelevine:
## Purpose

Add the word 'inactive' next to a location that is currently inactive.  
Because it should be more obvious when the location of an item is inactive.

## Approach

I repurposed some existing location variables that held the name to instead
hold the full location object, so that both name and isActive flag could be
accessed later.

## Pre-Merge Checklist

## Issues
https://issues.folio.org/browse/UIIN-1968


maccabeelevine:
Apparently I do need test coverage even though it's against the pattern in ViewHoldingsRecord.test.js, since otherwise this pushes the total coverage just below 80%.  I'll write something up but would still appreciate a look at the functional code to confirm it's the right approach.

maccabeelevine:
I added test coverage for one of the three scenarios.  Adding it for others would be possible but challenging since we'd need different holdings record stubs to test scenario 2 vs 3.  And probably overkill since there are no tests in ViewHoldingsRecord.test.js for any other data on the holdings record itself.  This just serves to get the coverage above 80%.

PR #1816, opened 2022-10-05
Disable some ESLint warnings

MikeTaylor:
I want a clean slate before I start work on UIIN-2197.

### DIFF
--- a/src/hooks/useInventoryBrowse/useInventoryBrowse.js
+++ b/src/hooks/useInventoryBrowse/useInventoryBrowse.js
@@ -86,8 +86,6 @@ const useInventoryBrowse = ({
 
   const updatePage = useCallback((_askAmount, _index, _firstIndex, direction) => {
     // TODO: implement prev-next pagination
-    // eslint-disable-next-line no-console
-    console.log('direction', direction);
   }, []);
 
   const {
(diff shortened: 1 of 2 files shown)


zburke (commented):
I _really_ appreciate the comments you added explaining why you're disabling lint where you do! But, I'm not a maintainer on this project any longer so disabling things isn't my call :) 

@mkuklis?

src/hooks/useInventoryBrowse/useInventoryBrowse.js
```diff
@@ -86,6 +86,7 @@ const useInventoryBrowse = ({
 
   const updatePage = useCallback((_askAmount, _index, _firstIndex, direction) => {
     // TODO: implement prev-next pagination
+    // eslint-disable-next-line no-console
     console.log('direction', direction);
```
mkuklis, line 90: @MikeTaylor let's just remove this `console.log`

MikeTaylor:
I am wary of removing actual code that someone added, presumably intentionally. So I'm happier removing the ESLint warning instead (especially given that there is an easy-to-grep `TODO` comment in place).

MikeTaylor:
On further overnight reflection, you're right. I removed the `console.log` because the TODO comment makes the relevant place in the code easy to find without it.

PR #1810, opened 2022-10-04
UIIN-2018: Add bound-with accordion to holdings

maccabeelevine:
## Purpose

Display a bound-with accordion on the holdings view, similar in concept to
[UIIN-1521](https://issues.folio.org/browse/UIIN-1521) which did the same on the
item view. 

## Approach

The accordion links to a search for each item by HRID, rather than to the item
view itself, because we do not have the instance UUID needed to construct the 
link to the item view and would need to add several additional queries on the 
bound-with items do so.

## Pre-Merge Checklist
Before merging this PR, please go through the following list and take appropriate actions.
- [X] I've added appropriate record to the CHANGELOG.md
- Does this PR meet or exceed the expected quality standards?
  - [ ] Code coverage on new code is 80% or greater
  - [ ] Duplications on new code is 3% or less
  - [ ] There are no major code smells or security issues
- Does this introduce breaking changes?
  - [ ] If any API-related changes - okapi interfaces and permissions are reviewed/changed correspondingly
  - [ ] There are no breaking changes in this PR.

## Issues
https://issues.folio.org/browse/UIIN-2018

PR #1828, opened 2022-10-13
UIIN-2212: Link from holdings view bound-with items

maccabeelevine:
## Purpose

The bound-with table on the holdings view lists the items bound-with 
that holding.  Each row should link to the item view for that item.

## Approach

#1810 previously solved this by linking to a search for the corresponding
item.  That was done because the data available in ViewHoldingsRecord.js
did not include the instance UUID necessary to create a direct link to the
item.  However that approach was rejected as bad UX (I agree) and reverted 
in #1823.

This second attempt extracts the accordion into a new HoldingBoundWith
component so that a separate query can be done to retrieve the parent
holdings record for each boundWithItem.  That holdings record contains
the instance UUID needed for the direct link.  This approach is patterned
after the existing HoldingAquisitions and HoldingReceivingHistory 
components.

## Pre-Merge Checklist

## Issues

https://issues.folio.org/browse/UIIN-2212


src/Holding/ViewHolding/HoldingBoundWith/fixtures.js
```diff
@@ -0,0 +1,7 @@
+// Disabling eslint below because the pattern is to name fixtures
```
maccabeelevine, line 1: I think this is correct, but please correct me if not.

zburke: Also, files named like `...test.js` will be excluded from the production build. Fixture data should be similarly excluded but we probably need to add a new rule for that, or move them into an already excluded directory such as `/test/` or `.../test/...`.

src/Holding/ViewHolding/HoldingBoundWith/HoldingBoundWith.js
```diff
@@ -0,0 +1,77 @@
...
+import { IntlConsumer } from '@folio/stripes/core';
+import { noValue } from '../../../constants';
+import { checkIfArrayIsEmpty } from '../../../utils';
+import useBoundWithHoldings from './useBoundWithHoldings';
+
+const HoldingBoundWith = ({ boundWithItems }) => {
+  const { isLoading, boundWithHoldings } = useBoundWithHoldings(boundWithItems);
+  const data = boundWithItems.records?.map(boundWithItem => ({
+    item: boundWithItem,
+    holdingsRecord: boundWithHoldings.find(
```
mkuklis, line 20: @maccabeelevine this could be simplified by first converting `boundWithHoldings` into a map via `keyBy` from `lodash`:

````js
const boundWithHoldingsMapById = keyBy(boundWithHoldings, 'id');
````

with that, you could turn it into:

````js
const data = boundWithItems.records?.map(boundWithItem => ({
  item: boundWithItem,
  holdings record: boundWithHoldings[boundWithItem.holdingsRecordId]
});
````
That should be a bit more performant since you won't need to loop via `boundWithHoldings` for every `boundWithItem`.

File: src/Holding/ViewHolding/HoldingBoundWith/useBoundWithHoldings.js [hunk 3]
```diff
@@ -0,0 +1,21 @@
+import { useQuery } from 'react-query';
+
+import { useOkapiKy } from '@folio/stripes/core';
+
+const useBoundWithHoldings = (boundWithItems) => {
+  const ky = useOkapiKy();
+
+  const queryKey = ['ui-inventory', 'bound-with-holdings', boundWithItems];
```
mkuklis, line 8: Please use `useNamespace` from stripes-core. Here are a couple of examples of how to use it:

https://github.com/folio-org/ui-inventory/search?q=useNamespace

src/Holding/ViewHolding/HoldingBoundWith/HoldingBoundWith.js
maccabeelevine: Good catch, I've made this change.

src/Holding/ViewHolding/HoldingBoundWith/useBoundWithHoldings.js
maccabeelevine: Thanks, that's a useful function.  I had patterned the query off the other two ViewHolding subcomponent helpers useHoldingOrderLines and useReceivingHistory which didn't use useNamepace().  I agree this is a better approach.

maccabeelevine:
@mkuklis I'd appreciate your advice.  There's a bug here where the link (to item view) fails with an error below if I navigate directly.  It succeeds if I refresh the error page, or if I open it in the first place with a right-click/open-in-new-tab, so the URL appears valid -- I'm guessing it's either a race condition or something needed with browser navigation history.

image

[Looks like firstItem is null here](https://github.com/folio-org/ui-inventory/blob/master/src/views/ItemView.js#L250) which means items.records[0] didn't exist.  I'm not clear when that resources.items object is supposed to be populated.

mkuklis (approved):
@maccabeelevine nice work! Thank you for the additional cleanup.

mkuklis:
@maccabeelevine that error you pointed out looks like some sort of regression. Let's merge your work and we can open another JIRA to cover the other bug.

mkuklis (approved):

PR #1832, opened 2022-10-20
UIIN-1969: Show inactive locations on item view

maccabeelevine:
## Purpose

Anywhere a location is specified on the item view, it should display the word 'Inactive' if 
that is the case.  This mirrors UIIN-1968 which did the same for the holdings view.

## Pre-Merge Checklist

## Issues
https://issues.folio.org/browse/UIIN-1969

### DIFF
--- a/src/views/ItemView.test.js
+++ b/src/views/ItemView.test.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import { BrowserRouter as Router } from 'react-router-dom';
 import { noop } from 'lodash';
+import { waitFor } from '@testing-library/react';
 
 import '../../test/jest/__mock__';
 
@@ -28,6 +29,7 @@ const resources = {
     records: [
       {
         permanentLocationId: 1,
+        temporaryLocationId: 'inactiveLocation',
       }
     ],
   },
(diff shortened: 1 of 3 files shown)


src/views/ItemView.test.js
```diff
@@ -122,5 +138,40 @@ describe('ItemView', () => {
       expect(document.querySelector('#item-list-bound-with-titles a.instanceHrid'))
         .toHaveAttribute('href', '/inventory/view/' + id);
     });
+
```
maccabeelevine, line 141: I would love feedback on whether all of these tests are necessary, or on the other hand insufficient.  They seem insufficient since to test fully we'd have to mock up maybe a dozen different item records to really test the presence (or absence) of the Inactive flag.  But on the other hand they seem unnecessary in that there are very few tests for any other read-only data on the item record, unless that itself is a problem we're trying to fix.

On the other hand if this is really what's intended I'd like to know that as well!

maccabeelevine:
Code coverage is under the threshold because I haven't tested the one feature (part of scenario 2) that adds the "Inactive" indication to ItemForm.  (The rest all involve ItemView.)  Please advise me on how to test the form.  Currently EditItem.test.js [mocks up the ItemForm](https://github.com/folio-org/ui-inventory/blob/master/src/Item/EditItem/EditItem.test.js#L16) as just a fake one-liner "ItemForm".  There's no item record at all to test against.

- Should that mock be (somehow) expanded? 
- Or should there be some new ItemForm.test.js file to mock up the form regardless of whether it's in Edit mode or Create/Duplicate?  All I'm testing is a static field value anyway.

Either way I'll have to do some learning but want to know which way to proceed.  Or should we not be testing this in the first place, or doing so as a separate Jira?

mkuklis (approved):

PR #1833, opened 2022-10-26
UIIN-2226: Fix displaying item via item route

mkuklis:
https://issues.folio.org/browse/UIIN-2226

The `items` resource is already used and loaded in the manifest defined in `ViewHoldingsRecord`:

https://github.com/folio-org/ui-inventory/blob/1e6b1760488e42d29509bc38da35366234cfb3d6/src/ViewHoldingsRecord.js#L83-L92

Unfortunately, this doesn't work well with the way stripes-connect has been implemented. The manifests are not scoped to instances (or in other words, they are not attached to components that defined them). They are placed in the global redux store. The keys in the manifest represent the parts of the redux store where the data is being stored.

When the same name is being used by multiple manifests defined across multiple components we occasionally may run into unexpected loading issues.

The problem here was that the `items` resource has been marked as loaded via ViewHoldingsRecord and it was incorrectly passing here:

https://github.com/folio-org/ui-inventory/blob/24d5a0d544905c91f6df34d8c3055d88efcb6eab/src/routes/ItemRoute.js#L185

before the items data has been loaded.


maccabeelevine:
That makes sense!  I also checked out the branch and the error appears gone.

maccabeelevine (approved):

PR #1839, opened 2022-10-31
UIIN-2227: Fix NPE on Add Item

maccabeelevine:
## Purpose
I caused an NPE on effectiveLocation in ItemForm.js with PR #1832.  The context was a change
to how an existing item's effectiveLocation is displayed, where it should never be null.  But
I failed to consider the AddItem context where it is not yet set.  This PR adds null checks.

## Pre-Merge Checklist

## Issues
https://issues.folio.org/browse/UIIN-2227

### DIFF
--- a/src/edit/items/ItemForm.js
+++ b/src/edit/items/ItemForm.js
@@ -266,7 +266,7 @@ class ItemForm extends React.Component {
 
     const labelLocation = get(holdingLocation, ['name'], '');
     const labelCallNumber = holdingsRecord.callNumber || '';
-    const effectiveLocation = locationsById[initialValues.effectiveLocation.id];
+    const effectiveLocation = locationsById[initialValues?.effectiveLocation?.id];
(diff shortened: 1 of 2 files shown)


src/edit/items/ItemForm.js
```diff
@@ -266,7 +266,7 @@ class ItemForm extends React.Component {
 
     const labelLocation = get(holdingLocation, ['name'], '');
     const labelCallNumber = holdingsRecord.callNumber || '';
-    const effectiveLocation = locationsById[initialValues.effectiveLocation.id];
+    const effectiveLocation = locationsById[initialValues?.effectiveLocation?.id];
```
maccabeelevine, line 269: The effectiveLocation null check is what matters here.  The initialValues check is probably not necessarily since it would have thrown an error earlier on line 216.  But I don't think it hurts.

zburke (approved):
LGTM. Thank you!

PR #1845, opened 2022-11-14
UIIN-1970: Show inactive locations on instance view

maccabeelevine:
## Purpose
Anywhere a location is specified on the instance view, it should display the word 'Inactive' if
that is the case. This mirrors UIIN-1968 and UIIN-1969 which did the same for the holdings
and item views respectively.

## Pre-Merge Checklist

## Issues
https://issues.folio.org/browse/UIIN-1970


maccabeelevine:
Thanks @mkuklis for the chat yesterday and taking a look at this.  I'm trying to write two test cases:

1. The first, stubbed, is to look at the accordion header for the word Inactive, which should appear since I've set the fake holding to use an inactive location.  But for some reason the variables in the ui-inventory.holdingsHeader are not being interpolated, so the test fails.
2. The next test should check the actual item in the accordion for its location, but that [existing stub](https://github.com/folio-org/ui-inventory/blob/e06dd6648ba8e2c7325343dece593db08f3ae859/src/Instance/HoldingsList/Holding/HoldingAccordion.test.js#L40) is empty.

mkuklis:
@maccabeelevine thank you for reaching out and pointing out these issues. 

The first issue related to translations should now work as expected. This is thanks to @zburke who pointed us to a more completed Harness in https://github.com/folio-org/stripes-data-transfer-components/blob/master/test/helpers/Harness.js

I'm not sure if I understand the second issue. Since the HoldingAccordion renders its `children` perhaps this may be out of the scope of this test. But please let me know if I understand what you had in mind.

maccabeelevine:
Thanks @mkuklis!  Confirmed that the first test passes now, thanks @zburke for the harness clue.

The second issue is to test the display of the inactive *item* location within the MCL within the accordion.  So yes it shouldn't be in any code related to HoldingAccordion, it should be in an ItemsList.test.js class which doesn't appear to exist.  If it's the correct approach I will make one, and use some version of the Item harness you pointed me to (but messing with it so that its location is inactive).  You explained that the lack of a test class here is probably a relic of dropping BigTest, so I get it.  Just want to make sure I'm not missing something, and that it's appropriate to create an ItemsList.test.js class just to test one particular column's edge case behavior.

mkuklis:
@maccabeelevine sounds good. 
I noticed a lot of other warnings when running these `jest` tests. I will spend a bit of time and try to clean them up.

mkuklis (approved):

maccabeelevine:
Hi @mkuklis I gave this a shot, it's not working yet so hoping for your advice:

> The second issue is to test the display of the inactive item location within the MCL within the accordion. So yes it shouldn't be in any code related to HoldingAccordion, it should be in an ItemsList.test.js class which doesn't appear to exist. If it's the correct approach I will make one, and use some version of the Item harness you pointed me to (but messing with it so that its location is inactive). You explained that the lack of a test class here is probably a relic of dropping BigTest, so I get it. Just want to make sure I'm not missing something, and that it's appropriate to create an ItemsList.test.js class just to test one particular column's edge case behavior.

I created ItemsList.test.js with a single test, and I'm using the existing items stub class.  The test fails with an error about "Invariant failed: Could not find required context" which [appears to be related to the drag-n-drop behavior](https://stackoverflow.com/questions/65851879/how-to-solve-invarient-failed-error-in-react).  I'm not clear how to stub out that behavior properly since my test doesn't care about it.

mkuklis:
@maccabeelevine we were missing another mock related to DND. Please try it again. That issue should not be present now.

maccabeelevine:
@mkuklis Thank you for that DND mock fix.  I was able to then fix the test case for the new ItemsList.test.js.  I think this is finally ready to go?

mkuklis (approved):
@maccabeelevine great work on this!

PR #1892, opened 2022-12-06
Fixes UIIN-2272: Do not fetch bound with data if the boundWithParts data is not present.

mkuklis:
https://issues.folio.org/browse/UIIN-2272

In cases when `holdingRecordIds` and `itemIds` were empty the query builder:

````js
const queryIds = `(${holdingRecordIds.join(' or ')})`;
````

would return `()` which later would convert to `true` with `Boolean(queryIds)` causing the query to execute and fail.

### DIFF
--- a/src/Holding/ViewHolding/HoldingBoundWith/useBoundWithHoldings.js
+++ b/src/Holding/ViewHolding/HoldingBoundWith/useBoundWithHoldings.js
@@ -11,7 +11,7 @@ const useBoundWithHoldings = (boundWithItems) => {
 
   const { data, isLoading } = useQuery(
     [namespace, queryIds],
-    () => ky.get(`holdings-storage/holdings?query=id=${queryIds}`).json(),
+    () => ky.get(`holdings-storage/holdings?query=id=(${queryIds})`).json(),
     { enabled: Boolean(queryIds) }
   );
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by maccabeelevine** [comment]
Ok that's embarrassing, that boolean definitely worked at one point but I refactored the query a few times and obviously missed it in the end.  LGTM.

**Review by maccabeelevine — APPROVED** [review]

File: src/Holding/ViewHolding/HoldingBoundWith/useBoundWithHoldings.js [hunk 1]
```diff
@@ -7,11 +7,11 @@ const useBoundWithHoldings = (boundWithItems) => {
   const [namespace] = useNamespace({ key: 'boundWithHoldings' });
 
   const holdingRecordIds = boundWithItems?.map(x => x.holdingsRecordId);
-  const queryIds = `(${holdingRecordIds.join(' or ')})`;
+  const queryIds = holdingRecordIds.join(' or ');
```
marcjohnson-kint, line 10: My interpretation of this change is that remove the quotes changes the produced string to be empty when there are no Ids. And this will then evaluate to false (due to JS's crazy type coercion).

How wrong is my understanding of this change?

mkuklis: @marcjohnson-kint yes, you are right the string will be empty but then the additional `(` `)` will be added creating `()` string. When that string is passed to `Boolean('()')` it will return `true` causing the `useQuery` to execute.

@maccabeelevine it's all good. I make many mistakes like that all the time too especially when refactoring/moving things around.

PR #1910, opened 2022-12-16
UIIN-1797 create Jest/RTL test for validateNameAndCode.js

cchitneedi:
Refs:
ui-inventory / UIIN-1797
URL: https://issues.folio.org/browse/UIIN-1797


KetineniM (approved):

KetineniM (approved):

ssandupatla (approved):

VSnehalatha (approved):

src/settings/validateNameAndCode.test.js
```diff
@@ -0,0 +1,19 @@
...
+    expect(JSON.stringify(errors.code)).toMatch(/ui-inventory.fillIn/i);
+    expect(JSON.stringify(errors.name)).toMatch(/ui-inventory.fillIn/i);
+  });
+  it('should return empty with name & code values present', () => {
+    const item = {
+      code: 'codeTest',
+      name: 'codeTest'
+    };
+    const errors = validateNameAndCode(item);
+    expect(errors).toBeDefined();
```
zburke, line 17: The stated purpose, "return empty with name & code values present" sounds great, but this test checks whether the return value is defined, not whether it is empty. Indeed, it needs to check that the returned object is empty in order to satisfy the question of whether errors are present.

src/settings/validateNameAndCode.test.js
```diff
@@ -0,0 +1,19 @@
+import '../../test/jest/__mock__';
+import validateNameAndCode from './validateNameAndCode';
+
+describe('ValidateNameAndCode', () => {
+  it('should return errors when code & name are empty', () => {
+    const item = {};
+    const errors = validateNameAndCode(item);
+    expect(JSON.stringify(errors.code)).toMatch(/ui-inventory.fillIn/i);
+    expect(JSON.stringify(errors.name)).toMatch(/ui-inventory.fillIn/i);
```
zburke, line 9: You may find it useful to [mock react-intl as in ui-users](https://github.com/folio-org/ui-users/blob/f96b1243f0b0a576b29d37d8cb742d455e9a6d13/test/jest/__mock__/intl.mock.js) so you can write tests that a little bit more concise:
```
expect(errors.code).toBe("ui-inventory.fillIn");
```

src/settings/validateNameAndCode.test.js
cchitneedi: Test name is changed to 'check return value is defined'

src/settings/validateNameAndCode.test.js
cchitneedi: We tried in this way but was getting error. So we kept as previous to match.

zburke (approved):

PR #1917, opened 2022-12-23
UIIN-1751 create Jest/RTL test for TargetProfiles.js

ssandupatla:
Refs:
Ticket - [UIIN-1751]
URL: https://issues.folio.org/browse/UIIN-1751
Command: npx jest --coverage TargetProfiles


cchitneedi (approved):

src/settings/TargetProfiles.test.js
```diff
@@ -0,0 +1,68 @@
...
+  </MemoryRouter>
+);
+const renderTargetProfilesSettings = () => renderWithIntl(
+  <TargetProfilesSetup />,
+  translationsProperties
+);
+describe('TargetProfiles Component', () => {
+  it('should take snapshot', () => {
+    const container = renderTargetProfilesSettings();
+    expect(container).toMatchSnapshot();
```
zburke, line 66: I'm not familiar with `toMatchSnapshot()`. The [snapshot documentation](https://jestjs.io/docs/snapshot-testing#snapshot-testing-with-jest) I found notes, 
> The snapshot artifact should be committed alongside code changes, and reviewed as part of your code review process. 

It looks like that didn't happen here. If you do go this route, please explain it in the PR where you introduce this pattern for the first time since it's a new test pattern. Please also explain why existing patterns are insufficient and/or why this one is better. You'll also need to investigate how we can exclude the `__snapshots__` directories from the artifacts we publish to NPM. To be clear, it is absolutely fine to introduce this new pattern, but you need to (a) do it correctly and (b) explain why you're doing it. 

Alternatively, see if there are other `EntryManager` tests. You might find it is sufficient to mock it as a simple component that echoes a few props (e.g. `label`) and your test can simply render and then check
```
expect(someProp).toBeInTheDocument();
```

zburke (changes requested):
`toMatchSnapshot()` needs some snapshot data to test against.

VSnehalatha (approved):

KetineniM (approved):

VSnehalatha (changes requested):
Render the component and then check for some props to be in the documents as Zak proposed

PR #1936, opened 2023-01-12
UIIN-2249: Get rid of "@folio/stripes-data-transfer-components" dependency

mariia-aloshyna:


mkuklis (approved):
Thank you @mariia-aloshyna!

OleksandrHladchenko1 (approved):

PR #1945, opened 2023-01-19
UIIN-2288: Align instance subjects and series fields with a new API breaking change

mkuklis:
https://issues.folio.org/browse/UIIN-2288

Align instance subjects and series fields with a new API breaking change.

subjects and series fields are now an array of objects:

````json
"properties": {
  "value": {
    "type": "string",
    "description": "Subject heading value"
  },
  "authorityId": { 
    "type": "string",
    "description": "UUID of authority record that controls a subject heading",
    "$ref": "uuid.json"
  }
}
````

### CODE REVIEW

File: package.json [hunk 1]
```diff
@@ -45,43 +45,43 @@
       }
     ],
     "okapiInterfaces": {
```
mkuklis, line 47: I ordered them alphabetically.

package.json
```diff
@@ -45,43 +45,43 @@
...
-      "identifier-types": "1.1",
-      "contributor-types": "2.0",
-      "contributor-name-types": "1.2",
-      "instance-types": "2.0",
+      "material-types": "2.0",
+      "modes-of-issuance": "1.0",
       "nature-of-content-terms": "1.0",
-      "instance-formats": "2.0",
-      "classification-types": "1.1",
+      "search": "0.7",
```
usavkov-epam, line 79: Looks like you forgot to update the interface to support version `1.0`

mkuklis: @usavkov-epam good catch! Thank you.

usavkov-epam (approved):

mkuklis:
@usavkov-epam do you know if this can be merged now or do we still need to wait?

usavkov-epam:
> @usavkov-epam do you know if this can be merged now or do we still need to wait?

AFAIK, we have to wait until the BE changes are merged and then we will get a ready signal from that command in slack I suppose

PR #1948, opened 2023-01-23
UIIN-1711 create Jest/RTL test for StatisticalCodeSettings.js

cchitneedi:
Refs: 
UIIN-1711 create Jest/RTL test for StatisticalCodeSettings.js
Reference: https://issues.folio.org/browse/UIIN-1711


cchitneedi:
Hi @mkuklis 

Regarding the following ticket could you please suggest me that how to define the "validateNameAndCode".  I was struck around line number 45 - 68  in the StatisticalCodeSettings.js file. I just written one test case (passed), but the major functionality covers in these lines. Can you please suggest.

Thanks & Regards,
Charan

mkuklis:
@cchitneedi I think the best would be to perhaps just mock `validateNameAndCode` function so you don't have to worry about its implementation.

cchitneedi:
Hi @mkuklis 
'ValidateNameAndCode' is mocked but there seems no improvement in coverage. Can you suggest how to reach the 'Validate' function.

Regards,
Charan

mkuklis:
@cchitneedi I think after `StatisticalCodeSettings` is rendered you will need to try to save it via test in order to trigger the validation function.

cchitneedi:
Hi @mkuklis ,
We have tried but no clue. Could you send any references similar, for our better understanding.

Thank You,
Charan

mkuklis:
@cchitneedi I have one more suggestion for you how to go about this:

1. Extract that `validate` function out of the component (you can place it above the component in the same file as a separate function) and export it with `export const validate = () {...}`.
2. Pass it to `ControlledVocab` here https://github.com/folio-org/ui-inventory/blob/master/src/settings/StatisticalCodeSettings.js#L138 as `validate` instead of `this.validate`.
3. Write a test just for the `validate` function by importing it directly in your test.

With this approach you will be able to test the validate and you won't need to fake the `ControlledVocab` which is not even part of ui-inventory

cchitneedi:
Hi @mkuklis ,
Thanks you for great suggestion!
Changes have been made in 'StatisticalCodeSettings.js' upon your suggestion, and added new test cases according to the changes in js file.

Thanks and Regards,
Charan

cchitneedi:
Hi @mkuklis,
I have updated and commited. Please review

Thanks and regards,
Charan

PR #1949, opened 2023-01-23
UIIN-1694 create Jest/RTL test for InstancesView.js

cchitneedi:
UIIN-1694 create Jest/RTL test for InstancesView.js

Reference: https://issues.folio.org/browse/UIIN-1694


cchitneedi:
Hi @mkuklis 

In InstancesView, I have attached the test file, where branches are not covered. I was unable to return values from prevProps.data and nextProps.data. Could you suggest me the best way to call return values or any.

Thanks and Regards,
Charan

mkuklis:
@cchitneedi I think the best approach to this would be to use `rerender` in a similar fashion to: https://github.com/folio-org/ui-users/pull/2287#issuecomment-1400547374

mkuklis (approved):

VSnehalatha (commented):
Approved

PR #1963, opened 2023-02-02
UIIN-2303: Bump stripes to 8.0.0 for Orchid/2023-R1

mkuklis:
https://issues.folio.org/browse/UIIN-2303

Bump stripes to 8.0.0 for Orchid/2023-R1

PR #1972, opened 2023-02-07
UIIN-2303 correctly bump stripes-connect to 8.1.0

zburke:
Whoops; missed stripes-connect in #1963

Refs [UIIN-2303](https://issues.folio.org/browse/UIIN-2303)


mkuklis (approved):
Oh crap thank you!

PR #1989, opened 2023-02-15
UIIN-2327: Disable deletion of a bound-with title

maccabeelevine:
## Purpose

According to the bound-with data model, an item should be bound-with linked
(via the bound-with parts API) to its own directly linked holdings record,
in addition to the one or more other holdings records it is bound with. 
Therefore it's incorrect to delete that directly-linked holdings record's 
relationship with the item record.

- Note that the backend API used for edits to the bound-with parts already enforces this relationship.  This PR only reflects that logic in the UI.
- As a corollary, note that even though the directly-linked bound-with title can't be removed via the UI, it will *automatically* be removed via that API call if it is the only remaining bound-with relationship on this item record.

## Approach
Disable the 'delete' action for a bound-with title on the Item Edit screen,
if that title (the indicated holdings record) is also directly linked to the item
as item.holdingsRecordId.

I enhanced FieldRow (used by RepeatableField) to support a function
to enable/disable the 'remove' action selectively per row, rather than 
universally.  For backwards compatibility it still supports and defaults
to expecting a boolean value.  I don't **think** this constitutes
a breaking change.

I made those changes to the ui-inventory version of these components, 
not the stripes-components version.  There are other UI differences between
 the identically-named components, and we wanted the UX to remain 
consistent between all of the ItemForm uses of RepeatableField.  Those 
differences include the layout of the delete button relative to the rest of 
the line, as well as the repetition of the field headers above each line.

## Pre-Merge Checklist

## Issues
https://issues.folio.org/browse/UIIN-2327


maccabeelevine:
@MikeTaylor I would appreciate an early review and in particular some help with the test case(s).  I wrote up what I think is a reasonable pair of tests, but it fails due to no such elements being found.  I encounter this a lot and @mkuklis always manages to find the race condition and add some more `await` style logic so it works.  I tried to do that myself but no luck.  I think Michal is on vacation this week and @CharlotteWhitt cares about this functionality, so it may or may not be fine to wait for him.  Thanks!

MikeTaylor:
Having discussed this with @maccabeelevine, we agreed that I'm not really competent to review this work. We may have to wait for @mkuklis.

maccabeelevine:
No worries @MikeTaylor, I ended up creating a new file entirely but got it to work.

src/components/RepeatableField/FieldRow.js
```diff
@@ -204,7 +204,10 @@ class FieldRow extends React.Component {
                         buttonStyle="link"
                         onClick={() => { this.handleRemove(fieldIndex, f); }}
                         aria-label={ariaLabel}
-                        disabled={!canDelete}
+                        disabled={
```
mkuklis, line 207: @maccabeelevine should we also protect for the cases when `canDelete` is not passed at all?

maccabeelevine: @mkuklis I think that's already taken care of by the defaultProps?

```
FieldRow.defaultProps = {
  canAdd: true,
  canEdit: true,
  canDelete: true,
};
```

mkuklis: Ah I missed the default props! This all looks good.

mkuklis (approved):

PR #1978, opened 2023-02-09
UIIN-2001: Delete bound-with parts

maccabeelevine:
## Purpose
Delete linked bound-with parts.  Represented in the UI as an edit to the item record.

## Approach
- Utilizes the new [PUT /inventory-storage/bound-withs](https://s3.amazonaws.com/foliodocs/api/mod-inventory-storage/p/bound-with-part.html#inventory_storage_bound_withs_put) API to replace the item's entire set of linked bound-with parts.  
- Although that API does both deletes and adds as needed, there's no UI functionality for Add yet so I've disabled the 'add' button.

### TODOS and Open Questions
- If the bound-with API call succeeds but then the Item update call fails, there is no rollback of the BW change.  I'm told there is no transactional behavior in the UI.

## Pre-Merge Checklist
Before merging this PR, please go through the following list and take appropriate actions.
- [x] I've added appropriate record to the CHANGELOG.md
- Does this PR meet or exceed the expected quality standards?
  - [x] Code coverage on new code is 80% or greater
  - [x] Duplications on new code is 3% or less
  - [x] There are no major code smells or security issues
- Does this introduce breaking changes?
  - [ ] If any API-related changes - okapi interfaces and permissions are reviewed/changed correspondingly
  - [X] There are no breaking changes in this PR.

## Issues
https://issues.folio.org/browse/UIIN-2001

PR #2034, opened 2023-03-03
UIIN-1998: Add bound-withs to item

maccabeelevine:
## Purpose
Add the UI capability to add bound-with parts to an Item, using a popup where the user inputs
Holdings Record HRIDs, following the "[simple edit solution](https://docs.google.com/presentation/d/1RxtJ6ZzqxkdITpWRWqGHF9Dz75cNbpktv4zgKM_SZ44/edit#slide=id.g1dbe3009864_0_0)".

Deleting bound-with parts was added in #1978 , so after this change, users can fully manage bound-with relations.

## Approach
- Since the popup is different than the usual 'add a row' mechanism of RepeatableField, I introduce a new parameter to fully hide its native "add" button, and I substitute my own.
- I encapsulate the FieldRows in a new BoundWithFieldRow component, to query and display the holdings and instance details related to the holdings record HRID entered by the user.  These aren't available in the item.boundWithParts yet since the updated item hasn't been saved.
- For the popup's form, I used a bare-bones React form and managed the (miniscule) state myself.  This is not hard since it always initializes as empty.  But if there's a better way via FinalForm, advice welcome.

## Pre-Merge Checklist

## Issues
https://issues.folio.org/browse/UIIN-1998


maccabeelevine:
I am not clear how to deal with the reported code smell.  There's no reasonable key I can imagine for those input fields on the popup, since they initialize to nothing, and there's always seven of them.

src/edit/items/BoundWithFieldRow.js
```diff
@@ -0,0 +1,53 @@
...
+  const holdingsRecordHrids = data.map(boundWithTitle => boundWithTitle.briefHoldingsRecord.hrid);
+  const { isLoading: isHoldingsLoading, holdingsRecords } = useHoldingsQueryByHrids(holdingsRecordHrids);
+
+  // Load the instance records matching the holdings' instanceIds.
+  const instanceIds = holdingsRecords.map(record => record.instanceId);
+  const instances = useInstancesQuery(instanceIds);
+  if (isHoldingsLoading || !instances.isSuccess) return <Loading size="large" />;
+
+  // Enrich the data render in the FieldRow
+  data.map(boundWithTitle => {
```
mkuklis, line 24: @maccabeelevine this is nitpicking but should we replace `map` with `forEach` since we are just looping here mutating `boundWithTitle` without returning any value at the end.

src/edit/items/BoundWithFieldRow.js
```diff
@@ -0,0 +1,53 @@
...
+    if (!boundWithTitle.briefHoldingsRecord?.id) {
+      const holdingsRecord = holdingsRecords.find(
```
mkuklis, line 26: @maccabeelevine I think you could avoid `find` here if the `keyBy` from lodash is used above (line 22 maybe):

`const holdingsRecordsByHrid = keyBy(holdingsRecords, 'hrid');`

Then here you could do:

`const holdingsRecord  = holdingsRecordsByHrid[boundWithTitle?.briefHoldingsRecord?.hrid]`

src/edit/items/BoundWithFieldRow.js
```diff
@@ -0,0 +1,53 @@
...
+        record => record.hrid === boundWithTitle.briefHoldingsRecord.hrid
+      );
+      boundWithTitle.briefHoldingsRecord.id = holdingsRecord.id;
+
+      const instance = instances.data?.instances?.find(x => x.id === holdingsRecord.instanceId);
```
mkuklis, line 31: `keyBy` could be also used here

mkuklis (commented):
@maccabeelevine while reviewing this, I realized that I am uncertain about the purpose of `BoundWithFieldRow`. From what I understand, it appears that all it does is to mutate the incoming data, which may be problematic and go against best practices. It may be more effective to ensure that the data being passed into the form via `initialValues` is already in the shape the form expects.

So ideally all additional data fetching, stitching and parsing should happen before the form is rendered (or re-rendered).

maccabeelevine:
Hi @mkuklis, I agree with all of the tweaks above and will do so.  Re the larger issue:

> @maccabeelevine while reviewing this, I realized that I am uncertain about the purpose of BoundWithFieldRow. From what I understand, it appears that all it does it to mutate the incoming data, which may be problematic and go against best practices. It may be more effective to ensure that the data being passed into the form via initialValues is already in the shape the form expects.
> 
> So ideally all additional data fetching, stitching and parsing should happen before the form is rendered (or re-rendered).

So yeah, this was difficult and you may be right, but it wasn't clear cut.  The data passed into the form via initialValues is a combination of any holdings that are already part of the item's bound-with parts (i.e. already in the DB), as well as any holdings IDs that were added via the BoundWithModal.  The Jira calls for displaying not only those user-entered HRIDS, but the instance HRID and title, which we have to load somewhere.  I thought it was reasonable to load them in the component that displays them, thus BoundWithFieldRow.  I'm happy to move that loading somewhere else that you suggest, per best practice.  One issue I ran into was that it uses two hooks to load that data, and there are React rules on where you can call hooks from.

src/edit/items/BoundWithFieldRow.js
maccabeelevine: Thanks, I have to get more familar with lodash -- this is much more efficient.

mkuklis:
I think ideally it would be great to load that data upfront in EditItem: 

https://github.com/folio-org/ui-inventory/blob/1b1d964cd5ba20c8a3d549aa533e2f8a08a7d640/src/Item/EditItem/EditItem.js#L28

And pass it via initialValues (you will have to mutate the item object):

https://github.com/folio-org/ui-inventory/blob/1b1d964cd5ba20c8a3d549aa533e2f8a08a7d640/src/Item/EditItem/EditItem.js#L113

The hooks may have to be refactored a bit to support multiple values at the same time. Perhaps a new hook could wrap both of them (essentially replacing `BoundWithFieldRow` component)?

mkuklis (approved):
I met with @maccabeelevine and we reviewed this together. The thing which I described in my previous comment maybe hard to achieve because the same lookup has to happen for the newly added row. 

Ideally what should happen here (as pointed by @maccabeelevine) is to introduce a new plugin called `ui-find-hodlings-record-plugin` which would allow for searching and returning holding records. 

It's something which can be improved in the future but we should go with this work as a v1 so we don't loose the momentum. Nice work @maccabeelevine!

PR #2060, opened 2023-03-17
UIIN-1609 create Jest/RTL test for instanceFilterRenderer.js

cchitneedi:
Refs:
UIIN-1609
create Jest/RTL test for instanceFilterRenderer.js

URL: https://issues.folio.org/browse/UIIN-1609


ssandupatla (approved):

src/components/InstanceFilters/instanceFilterRenderer.test.js
```diff
@@ -0,0 +1,47 @@
...
+    expect(screen.getByText('Effective location (item)')).toBeInTheDocument();
+  });
+
+  it('effectiveLocation field', () => {
+    expect(screen.getByText('effectiveLocation-field')).toBeInTheDocument();
+  });
+
+  it('language-field', () => {
+    expect(screen.getByText('language-field')).toBeInTheDocument();
+  });
```
zburke, line 46: These are assertions on the internals of `<InstanceFilters>`. Instead, perhaps that should be mocked so you can assert that it receives the expected props.

zburke (changes requested):

src/components/InstanceFilters/instanceFilterRenderer.test.js
cchitneedi: Hi @zburke , We have mentioned what was present in DOM for writing test cases, could you please elaborate what needs to be mocked in this case so that we can solve the issue ASAP. 

Thanks and Regards,
Charan

cchitneedi: Hi @zburke,
File has been updated as suggested.

Thanks and Regards,
Charan

ryandberger (approved):

VSnehalatha (approved):

PR #2070, opened 2023-03-21
remove lokalise cruft

zburke:
The `translations/{repo}` directory appears to be a glitch from a bad lokalise merge. The files are gigantic (e.g. 867639 instead of 45610 for en-US) and appear to contain values from several different repositories.


dltj (approved):
Whoa...thanks. Not sure what happened 2 months ago.

PR #2090, opened 2023-03-29
UIIN-1671 create JEST/RTL for ItemsListRow.js

ssandupatla:
REFS: UIIN-1671 create JEST/RTL for ItemsListRow.js
URL: https://issues.folio.org/browse/UIIN-1671

### DIFF
--- a/src/Instance/ItemsList/ItemsListRow.test.js
+++ b/src/Instance/ItemsList/ItemsListRow.test.js
@@ -1,0 +1,79 @@
+import React from 'react';
+import { screen } from '@testing-library/react';
+
+import '../../../test/jest/__mock__/currencyData.mock';
+import '../../../test/jest/__mock__/stripesConfig.mock';
+import '../../../test/jest/__mock__/stripesCore.mock';
+import { renderWithIntl } from '../../../test/jest/helpers';
+
+import ItemsListRow from './ItemsListRow';
+
+jest.mock('react-beautiful-dnd', () => ({
+  Draggable: jest.fn().mockImplementation(({ children }) => children({
+    draggableProps: {
+      style: {},
+    },
+    innerRef: jest.fn(),
+  },
+  {
+    isDragging: false
+  }))
+    .mockImplementationOnce(({ children }) => children({
+      draggableProps: {
+        style: {},
+      },
+      innerRef: jest.fn(),
+    },
+    {
+      isDragging: true
+    })),
+}));
+
+const mockisItemsDragSelected = jest.fn();
+const defaultProps = {
+  rowClass: [
+    'RowClass 1',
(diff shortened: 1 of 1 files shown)


cchitneedi (approved):

src/Instance/ItemsList/ItemsListRow.test.js
```diff
@@ -0,0 +1,79 @@
...
+    container = document.createElement('div');
+    container.setAttribute('id', 'ModuleContainer');
+    document.body.appendChild(container);
+  }
+  renderWithIntl(<ItemsListRow {...props} />);
+};
+
+
+describe('ItemsListRow', () => {
+  it('new mo should render isDragging of snapshot is true', () => {
```
zburke, line 71: I don't understand what "new mo" means; is this a typo? I don't understand how the description "isDragging ... is true" relates to the expect clause

src/Instance/ItemsList/ItemsListRow.test.js
```diff
@@ -0,0 +1,79 @@
...
+    renderItemsListRow(defaultProps);
+    expect(screen.queryByText('ui-inventory.moveItems.move.items.count')).toBeInTheDocument();
+  });
+  it('child should render isDragging of snapshot is false', () => {
+    renderItemsListRow(defaultProps);
+    expect(screen.getByRole('row', { name: 'cell 1cell 2cell 3' })).toBeInTheDocument();
```
zburke, line 77: I don't understand how the description "isDragging ... is false" relates to the expect clause.

zburke (changes requested):
The two test descriptions suggest testing opposite conditions (isDragging should be true, then isDragging should be false) and use the two different mock implementations of react-beautiful-dnd. OK, that part makes sense. But then you should be checking the same condition in both places, i.e. assert that something is true in the first test, then false in the second test. Alternatively, if you can't leverage the same element for both tests, then include both `expect` clauses in both tests and check that their conditions flip, e.g.
```
it('test 1', () => {
  expect(first condition).toBeInTheDocument();
  expect(second condition).not.toBeInTheDocument();
});

it('test 2', () => {
  expect(first condition).not.toBeInTheDocument();
  expect(second condition).toBeInTheDocument();
});
```

src/Instance/ItemsList/ItemsListRow.test.js
ssandupatla: We have cleared all typos. Changes made as per your suggestions.

src/Instance/ItemsList/ItemsListRow.test.js
ssandupatla: We have cleared all typos. Changes made as per your suggestions. Updated test description.

PR #2091, opened 2023-03-29
UIIN-1766 create JEST/RTL for InstanceNotesList.js

ssandupatla:
REFS: UIIN-1766 create JEST/RTL for InstanceNotesList.js
URL: https://issues.folio.org/browse/UIIN-1766

### DIFF
--- a/src/Instance/InstanceDetails/InstanceNotesView/InstanceNotesList.test.js
+++ b/src/Instance/InstanceDetails/InstanceNotesView/InstanceNotesList.test.js
@@ -1,0 +1,42 @@
+import React from 'react';
+import { BrowserRouter as Router } from 'react-router-dom';
+
+import '../../../../test/jest/__mock__';
+import renderWithIntl from '../../../../test/jest/helpers/renderWithIntl';
+
+import InstanceNotesList from './InstanceNotesList';
+
+const props1 = {
+  id: 'InstanceNotesListID',
+  notesType: 'test-notes',
+  notes: [{ staffOnly: true, note: 'Note 1' }],
+};
+
+const noValueProps = {
+  id: 'test-id',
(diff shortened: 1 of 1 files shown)


cchitneedi (approved):

VSnehalatha (approved):

src/Instance/InstanceDetails/InstanceNotesView/InstanceNotesList.test.js
```diff
@@ -0,0 +1,42 @@
...
+    expect(getByText('ui-inventory.staffOnly')).toBeInTheDocument();
+    expect(getByText('test-notes')).toBeInTheDocument();
+    expect(getByText('ui-inventory.yes')).toBeInTheDocument();
+    expect(getByText('Note 1')).toBeInTheDocument();
+    expect(getByText('stripes-components.endOfList')).toBeInTheDocument();
+  });
+  it('should render the noValue component when notes is empty', () => {
+    const { getByText } = renderInstanceNotesList(noValueProps);
+    expect(getByText('test-notes')).toBeInTheDocument();
+  });
```
zburke, line 41: I don't understand how this test is related to the `noValue` component. Rather, it validates that the note's type is used as a column header, which is the same thing tested on line 33. From a coverage point of view, you're _running_ the `noValue` code here, but not asserting on it. Effectively that throws away the value of the test.

zburke (changes requested):
If you comment out all the `expect` clauses in these tests, you'll still see 100% coverage. Running the code isn't the same as asserting on it. To test whether it will "render the noValue component when notes is empty", assert on the absence of a (mocked) noValue when notes is populated and on the presence of noValue when notes is empty.

PR #2104, opened 2023-04-03
Release v9.4.5

OleksandrHladchenko1:


mkuklis:
@OleksandrHladchenko1 something doesn't look right here. It looks like the branch `b9.4` meant for Orchid has been merged into master. I think we need to revert this.

mkuklis:
@OleksandrHladchenko1 actually never mind. I think I understand what happened here now. This is a bit different from how we manage this situation and I was worry that all the commits from b9.4 landed on master but it looks like you just wanted to get the changelog in place.

mkuklis:
@OleksandrHladchenko1 I actually do have one question about it. It looks like you merged master into `b9.4`: https://github.com/folio-org/ui-inventory/commit/2cda09e11ccfba1b30107e2283bd0d30407034ea

b9.4 was a branch meant to be used for Orchid releases but it now includes the work from master which means some work meant only for Poppy for example:

https://issues.folio.org/browse/UIIN-2356

I'm curious what commit was used to release `v9.4.5`.

Also I think `b9.4` should be probably removed now since it contains master because if you need to release another version `v9.4.6` in the future you will want to make sure the work for Poppy is not included.

OleksandrHladchenko1:
Hi @mkuklis. My apologies, I merged master into `b9.4` by mistake. I just tried to release [this ticket](https://issues.folio.org/browse/UIIN-2343).

mkuklis:
@OleksandrHladchenko1 no worries. Let's remove that b9.4 then and maybe create a new branch out of the commit from 9.4.5 release for the future Orchid releases.

zburke:
@OleksandrHladchenko1, I agree with @mkuklis that `master` should not have been merged into `b9.4` (this means we can no longer use the `b9.4` (Orchid) branch for additional patch releases since it contains work from `master` that is intended for Poppy). It was also unnecessary to merge `b9.4` into master as all those commits should already part of master. 

Our typical release process is outlined in [folio-org/stripes/doc/release-procedure.md#patch-release-procedures](https://github.com/folio-org/stripes/blob/8dafa554624a6d84e3970c3cdc309b66e0477583/doc/release-procedure.md#patch-release-procedures). That documentation is a bit outdated (note references to Jenkins) but the important parts remain unchanged. It's up to you whether to use an extra branch and PR for the release commit or if you want to just commit directly on the branch. The advantage of a PR is that it runs the tests, giving you more confidence that the release is sound. The advantage of commiting directly on the branch is that's a less hassle.

PR #2134, opened 2023-04-25
UIIN-2405: Holdings view source: Print button not visible with "View MARC holdings record" permission

przemyslawturek:
Holdings view source: Print button not visible with "View MARC holdings record" permission

## Issue
[UIIN-2405](https://issues.folio.org/browse/UIIN-2405)


src/components/ViewSource/ViewSource.js
```diff
@@ -32,7 +32,7 @@ const ViewSource = ({
   const stripes = useStripes();
 
   const isPrintBibAvailable = !isHoldingsRecord && stripes.hasPerm('ui-quick-marc.quick-marc-editor.view');
-  const isPrintHoldingsAvailable = isHoldingsRecord && stripes.hasPerm('ui-quick-marc.quick-marc-holdings-editor.all');
+  const isPrintHoldingsAvailable = isHoldingsRecord && (stripes.hasPerm('ui-quick-marc.quick-marc-holdings-editor.all') || stripes.hasPerm('ui-quick-marc.quick-marc-holdings-editor.view'));
```
BogdanDenis, line 35: It's ok to just check for `ui-quick-marc.quick-marc-holdings-editor.view`
Because if user has `ui-quick-marc.quick-marc-holdings-editor.all` - they will get the View permission automatically

BogdanDenis (changes requested):

src/components/ViewSource/ViewSource.js
przemyslawturek: I think there is a logical problem with permission. When user has ui-quick-marc.quick-marc-holdings-editor.all he also get ui-quick-marc.quick-marc-holdings-editor.view perm as subperms but not opposite. When has only view permission he dosn't have ui-quick-marc.quick-marc-holdings-editor.all - and that's why isPrintHoldingsAvailable is false

BogdanDenis (approved):

mariia-aloshyna (approved):

PR #2152, opened 2023-05-16
UIIN-2427: Import testing-library deps from @folio/jest-config-stripes

mkuklis:
https://issues.folio.org/browse/UIIN-2427

Consume @testing-library deps via `@folio/jest-config-stripes`.


zburke (approved):
LGTM!

PR #2157, opened 2023-05-23
UIIN-2365 Remove unnecessary override styles for TextLink

JohnC-80:
Problem
ui-inventory unnecessarily overrides styles that are provided by stripes-components. At the time the style was implemented, the styles were necessary since `TextLink` styles were being overriden by `MCL`'s own styles. This was resolved in https://github.com/folio-org/stripes-components/pull/2010

Screenshots do the talking:
image


zburke (changes requested):
The fix in stripes-components won't land until `v11.1.0`, which will correspond to stripes `v8.1.0`, so I think this change needs to accompanied by a version bump in `package.json`.

zburke (approved):
Thanks! LGTM.

Dmytro-Melnyshyn (approved):

PR #2171, opened 2023-06-05
UIIN-2443: Rename hrid qindex for item to avoid collisions with holdings ans instances

mkuklis:
https://issues.folio.org/browse/UIIN-2443

The issue here was related to the work done in: https://issues.folio.org/browse/UIIN-2036.

In UIIN-2036 the idea was to be able to automatically open an item details view if the search was performed on one of these indexes:

````js
  'items.barcode': 'barcode==%{query}',
  isbn: 'isbn==%{query}',
  issn: 'issn==%{query}',
  hrid: 'hrid==%{query}',
  iid: 'id==%{query}'
````

and the returned instance had a single item attached to it.

The problem with that approach is that the `hrid` index name has been the same for instances, holdings and items. 
If somebody searched for holdings record hrid there was no way to detect it and the code alway assumed that the search for item hrid was performed causing to open the item details screen (even though the intention was to search for the holdings record).

This PR renames the hrid index for the item so it's clear that somebody is searching by item hrid (and not instance or holdings record).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -147,6 +147,7 @@
 * Moved the print button from QuickMarkView, into the ViewSource. Due to the folio-org/ui-quick-marc#468. Refs UIIN-2324.
 * Change title for the print popup. Refs UIIN-2329.
 * Move @testing-library/* to dev-deps. Refs UIIN-2309.
+* Rename `hrid` qindex for item to avoid collisions with holdings and instances. Fixes UIIN-2443.
 
 ## [9.2.0](https://github.com/folio-org/ui-inventory/tree/v9.2.0) (2022-10-27)
 [Full Changelog](https://github.com/folio-org/ui-inventory/compare/v9.1.0...v9.2.0)
--- a/src/constants.js
+++ b/src/constants.js
@@ -434,6 +434,6 @@ export const SINGLE_ITEM_QUERY_TEMPLATES = {
   'items.barcode': 'barcode==%{query}',
   isbn: 'isbn==%{query}',
   issn: 'issn==%{query}',
-  hrid: 'hrid==%{query}',
+  itemHrid: 'hrid==%{query}',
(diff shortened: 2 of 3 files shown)


MikeTaylor (commented):
But does the backend understand the `itemHrid` index?

If not, I would have expected to see something in the item-searching make-query function that transforms `itemHrid`.

mkuklis:
@MikeTaylor `itemHrid` is just a key referenced on the front end.

mkuklis:
@MikeTaylor as you rightly pointed out the mapping is happening here:

https://github.com/folio-org/ui-inventory/blob/3eb71003ae57608d4982d0ce7f264b08c290fd26/src/utils.js#L764

the key in that map (`qindex`) is coming from the `value` in: 

https://github.com/folio-org/ui-inventory/pull/2171/files#diff-03423e533e7101fe103a71bd392486b92db5623a15d76ce0bbd3a96c39668017R251

That value (`qindex`) is being picked when the user switches the search option in the UI:

qindex

After this happens the URL will be updated with the correct `qindex`:

qindex-in-url

Now when the user performs the search there will be extra check for that index in:

https://github.com/folio-org/ui-inventory/blob/3eb71003ae57608d4982d0ce7f264b08c290fd26/src/components/InstancesList/InstancesList.js#L944

If the qindex belongs to the `SINGLE_ITEM_QUERY_TEMPLATES` map it will be mapped to the correct template (used to communicate with the mod-search).

MikeTaylor:
Argh! I can't believe my own stupidity. I did enough Code Delving to realise that you needed to make a change in `itemIndexes`, but somehow overlooked that your pull request _already included this change_.

Jeez.

I will go and lie down.

MikeTaylor (approved):
Do it!

mkuklis:
@MikeTaylor no honestly this is kind of hard to follow and your questions were all good!

PR #2190, opened 2023-06-28
UIIN-1392: When removing the dead icons in the Item record, then the displayed text looks inconsistent

mariia-aloshyna:
Added translations for item statuses in lowercase format for English local so for German and other languages it should look correct now.


mariia-aloshyna:
Hi @zburke What do you think about this? I added translations for item statuses and material types in lowercase format for English local so for German and other languages it should look correct now

zburke (changes requested):
This works for item-statuses since that's a fixed list. Because [item material types are dynamic](https://folio-snapshot.dev.folio.org/settings/inventory/materialtypes) we should not specify translation for them.

src/components/ItemViewSubheader/LinkedInstanceDetails.js
```diff
@@ -0,0 +1,56 @@
...
+const LinkedInstanceDetails = ({
+  item,
+  instance,
+}) => {
+  const boundWithCount = item?.boundWithTitles?.length;
+  const instancePublicationCount = instance.publication?.length;
+  const linkedInstanceTitle = (
+    <Link to={`/inventory/view/${instance.id}`}>
+      {` ${instance.title} `}
+    </Link>
```
zburke, line 17: Is this correct or do we need sth like
```
<>{' '}<Link>{title}</Link>{' '}</>
```
so the whitespace surrounding the title is not part of the link? I don't know exactly how this will render.

zburke (approved):
LGTM, though please make sure the link formatting is correct.

src/components/ItemViewSubheader/LinkedInstanceDetails.js
```diff
@@ -0,0 +1,56 @@
...
+import PropTypes from 'prop-types';
+
+import css from '../../View.css';
+
+const LinkedInstanceDetails = ({
+  item,
+  instance,
+}) => {
+  const boundWithCount = item?.boundWithTitles?.length;
+  const instancePublicationCount = instance.publication?.length;
```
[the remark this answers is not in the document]
mariia-aloshyna: good point

src/components/ItemViewSubheader/LinkedInstanceDetails.js
mariia-aloshyna: I removed this whitespace surrounding the title actually because the translation looks like this `"Instance: {title}."` so it renders correctly. Just copied this part of the code when refactored

zburke (approved):
LGTM

OleksandrHladchenko1 (approved):

PR #2191, opened 2023-06-29
UIIN-2467: Decrease rerenders for TargetProfileDetail component

OleksandrHladchenko1:
## Purpose

* When select target profile, [error](https://legacy.reactjs.org/docs/error-decoder.html/?invariant=185) appears 

## Approach

* Made `TargetProfileDetail` component as functional.
* Created custom hooks `useAllowedJobProfiles` and `useDefaultJobProfile` to fetch job profiles, instead of manifest.
* Fixed and added unit tests.

## Pre-Merge Checklist


src/settings/TargetProfileDetail.js
```diff
@@ -133,10 +47,10 @@ class TargetProfileDetail extends React.Component {
       id: jobProfile?.id,
       name: jobProfile?.name,
     }));
-  }
+  };
 
-  getJobProfilesFormatter = defaultProfileId => ({
-    [JOB_PROFILES_COLUMNS_NAME.ID]: ({ name, id }) => (
+  const getJobProfilesFormatter = defaultProfileId => ({
+    [JOB_PROFILES_COLUMNS_NAME.ID]: ({ name, id }) => ( // eslint-disable-line react/prop-types
```
zburke, line 53: Lint is telling you, "Hey, this is actually a functional component and components need prop-types" so instead of shutting it off, just follow its (somewhat oblique) advice: lift this function up and out of the body of `TargetProfileDetail`, or maybe even into its own file, and treat it like the real component that it is.

src/settings/TargetProfileDetail.js
```diff
@@ -147,148 +61,153 @@ class TargetProfileDetail extends React.Component {
         <span>{`(${id})`}</span>
       </>
     ),
-    [JOB_PROFILES_COLUMNS_NAME.IS_DEFAULT]: ({ id }) => {
+    [JOB_PROFILES_COLUMNS_NAME.IS_DEFAULT]: ({ id }) => { // eslint-disable-line react/prop-types
```
zburke, line 64: As above.

src/settings/TargetProfileDetail.test.js
```diff
@@ -142,10 +144,10 @@ describe('TargetProfileDetail', () => {
   });
 
   it('names of job profiles should be displayed as a hotlink', () => {
-    const { getByText } = renderTargetProfileDetail();
+    const { getAllByText } = renderTargetProfileDetail();
 
-    expect(getByText('create name').href).toContain('/settings/data-import/job-profiles/view/createJobProfileTestId');
-    expect(getByText('update name').href).toContain('/settings/data-import/job-profiles/view/updateJobProfileTestId');
+    expect(getAllByText('Test name 1')[0].href).toContain('/settings/data-import/job-profiles/view/testId1');
+    expect(getAllByText('Test name 1')[1].href).toContain('/settings/data-import/job-profiles/view/testId1');
```
zburke, line 150: It feels like we're losing detail here: we used validate that the "create..." link corresponded to a "create..." url, ditto for "update", and now we're validating that some array elements (who knows if/how 0 and 1 are supposed to be the same or different) match. 

Is it possible to keep the level of detail we had before, or am I misunderstanding?

zburke (changes requested):
Wow, it's almost shocking how much cleaner the `TargetProfileDetail` component is when the resource handling is moved out to hooks! Thank you! This is really nice. 

That said, I don't have any understanding of what problem the actual bug was. The PR only describes the symptoms, never their source, and I was never able to reproduce the behavior in the ticket. If you don't understand it either, that's fine, but at least describe the behavior you saw ("saw this error message, saw this stacktrace") and why you made the changes you did ("suspect a problem with X; replaced the manifest with hooks to avoid X"). Also, while `TargetProfileDetail` got a lot simpler (yay!) the directory structure of the new hooks is unnecessarily complicated. 

With a little more cleanup work, this PR will move from good to great.

src/settings/TargetProfileDetail.js
OleksandrHladchenko1: Let's leave it as is, but remove destructuring like [here](https://github.com/folio-org/ui-inventory/blob/master/src/Instance/InstanceDetails/InstanceTitleData/TitleSeriesStatements.js#L27) and [here](https://github.com/folio-org/ui-inventory/blob/ded70e7a371366947b4909706220c6fc1afae32c/src/Instance/InstanceDetails/InstanceTitleData/AlternativeTitlesList.js#L53)

zburke (approved):
I still wish we had a better understanding of the original problem so we could more strongly state how/why this refactor solves it ... but I can't argue at all with how much cleaner and easier to understand this code is. Nice work!

src/common/hooks/useAllowedJobProfiles.js
```diff
@@ -0,0 +1,35 @@
...
+  const DATA_TYPE_MARC_QUERY = 'dataType==("MARC")';
+
+  const ids = allowedJobProfileIds.join(' or ');
+  const allowedIdsQuery = `id==(${ids})`;
+  const path = `${DATA_IMPORT_JOB_PROFILES_ROUTE}?limit=${LIMIT_MAX}&query=${DATA_TYPE_MARC_QUERY} and ${allowedIdsQuery} sortBy name`;
+
+  const { isLoading, data: allowedJobProfiles = {} } = useQuery({
+    queryKey: [namespace, allowedJobProfileIds],
+    queryFn: () => ky.get(path).json(),
+    enabled: !!allowedJobProfileIds,
```
mariia-aloshyna, line 26: This array can be empty but you still will get 'true' so you need to check that it's not empty or just check your 'ids' string

PR #2193, opened 2023-06-30
UIIN-2478: Fix useBoundWithHoldings for 100+ items

maccabeelevine:
## Purpose
useBoundWithHoldings calls GET holdings-storage/holdings with a query of a list of holdings UUIDs combined with 'or'.  This works but fails when the number of holdings UUIDs causes the URL to exceed the length limit. 

## Approach
There are several CQL "chunking" solutions via react-query useQuer**ies**()  across the UI modules:
https://github.com/search?q=org%3Afolio-org%20useQueries&type=code

The most recent, easiest to use, and (looks to me) most robust is in stripes-erm-components:
https://github.com/folio-org/stripes-erm-components/blob/master/lib/hooks/useChunkedCQLFetch.js

ui-inventory currently doesn't appear to use stripes-erm-components (not surprising, it's not part of ERM).  Since I'm solving this as part of a potential release patch, I didn't want to add a new dependency, so I've temporarily COPIED THE FILE WHOLE into ui-inventory.  I would suggest "promoting"? the hook from stripes-erm-components into stripes-components and then importing it from there.

#### TODOS and Open Questions
- [ ] Remove the copied useChunkedCQLFetch.js, import from somewhere.

## Pre-Merge Checklist

## Issue
https://issues.folio.org/browse/UIIN-2478

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@
 * Adjust the Instance Edit screen header. Refs UIIN-2437.
 * Inventory: Retain Search/Browse query/options/filter and facet selections UNLESS user resets/clear selections. Refs UIIN-2433.
 * Statistical Code dropdown "contains" type ahead functionality needed (Instance/Holdings/Items). Refs UIIN-2466.
+* Fix problem with a large number of items on a single holdings record. Refs UIIN-2478.
 
 ## [9.4.7](https://github.com/folio-org/ui-inventory/tree/v9.4.7) (2023-06-27)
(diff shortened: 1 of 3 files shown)


maccabeelevine:
Coverage on useBoundWithHoldings.js is still over 80%.  The failure (barely, 79.6%) is due to the copied-from-ERM hook, which I expect we'll be deleting once we import the existing version of it as I suggested above.  So I think it's ok as-is, but up to the reviewers.

mariia-aloshyna (approved):

PR #2213, opened 2023-07-19
UIIN-2442 Add Shared icon to inventory instance results

BogdanDenis:
## Description
Add Shared icon to inventory instance results

## Screenshots
image

## Issues
[UIIN-2442](https://issues.folio.org/browse/UIIN-2442)


src/components/InstancesList/InstancesList.js
```diff
@@ -1077,39 +1090,53 @@ class InstancesList extends React.Component {
         title,
         discoverySuppress,
         isBoundWith,
+        shared,
```
mariia-aloshyna, line 1093: Where does this property come from?

BogdanDenis: From `contentData` array. It's a new property of `instance`

Dmytro-Melnyshyn (approved):

src/components/InstancesList/InstancesList.js
mariia-aloshyna: Good to know, thank you @BogdanDenis !

PR #2233, opened 2023-08-14
UIIN-2508 Updated version of ui-plugin-find-instance

BogdanDenis:
## Description
Update version of `ui-plugin-find-instance` after uprading it to React 18

## Related PRs
https://github.com/folio-org/ui-plugin-find-instance/pull/172

### DIFF
--- a/package.json
+++ b/package.json
@@ -912,7 +912,7 @@
   },
   "optionalDependencies": {
     "@folio/plugin-create-inventory-records": "^3.0.0",
-    "@folio/plugin-find-instance": "^6.0.0"
+    "@folio/plugin-find-instance": "^7.0.0"
   },
   "resolutions": {
     "moment": "~2.29.4"


package.json
```diff
@@ -912,7 +912,7 @@
   },
   "optionalDependencies": {
     "@folio/plugin-create-inventory-records": "^3.0.0",
-    "@folio/plugin-find-instance": "^6.0.0"
+    "@folio/plugin-find-instance": "^7.0.0"
```
Dmytro-Melnyshyn, line 915: Please update the changelog.

Dmytro-Melnyshyn (approved):

package.json
BogdanDenis: I made this change as a second part of UIIN-2508 and it has an entry already

mariia-aloshyna (approved):

PR #2235, opened 2023-08-15
UIIN-2529: Change tenant id to central when opening details of shadow instance

mariia-aloshyna:
## Purpose

To change X-Okapi-Tenant header to the central tenant id to be able to display the shared instance for a Local member tenant.

## Approach

As far as we cannot get the `source` value of an instance to indicate whether it is shared or not I fetch by `search/instance/` and get the `shared` prop from here, so if you have a better idea of how to avoid additional request I will be glad to discuss.

**NOTE** - move `useTenantKy` to `stripes-core` and re-use. For now just duplicated from `ui-consortia-settings` module

## Refs

https://issues.folio.org/browse/UIIN-2529

### DIFF
--- a/src/ViewInstanceWrapper.js
+++ b/src/ViewInstanceWrapper.js
@@ -1,6 +1,5 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { omit } from 'lodash';
 
 import { checkIfUserInMemberTenant } from '@folio/stripes/core';
 import { withTags } from '@folio/stripes/smart-components';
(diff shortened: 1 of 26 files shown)


BogdanDenis (approved):

OleksandrHladchenko1 (approved):

src/ViewInstanceWrapper.js
```diff
@@ -1,26 +1,34 @@
...
   } = props;
 
-  const instance = resources.instance?.records?.[0];
-  const selectedInstance = instance?.id === id ? instance : null;
   const userId = stripes?.user?.user?.id;
+  const tenantId = stripes.okapi.tenant;
   const centralTenantId = stripes.user.user?.consortium?.centralTenantId;
 
+  const { instance: selectedInstance } = useSearchInstanceByIdQuery(id);
+  const { instance } = useInstanceQuery(
```
Dmytro-Melnyshyn, line 26: Please extend the `checkIfSharedInstance` function, now we also need to check if `intance.shared`, because `source` will not always be with consortia prefix for member tenant.

mariia-aloshyna: Hi @Dmytro-Melnyshyn to extend the `checkIfSharedInstance` function it's needed to make an additional call everywhere where it's used already and pass the `instance` property containing the `shared` prop, because now all the components that use this function are fetching instance by `inventory/instances/:{id}` endpoint, and here I used `search/instances/query=id=={id}`

Dmytro-Melnyshyn (approved):

Dmytro-Melnyshyn:
@mariia-aloshyna I think all requests should be with the central tenant id in X-Okapi-Tenant when we open the instance detail view of shared record.

PR #2240, opened 2023-08-21
UIIN-2530: Enable pagination in the Instance's Holding Item list

OleksandrHladchenko1:
## Purpose

* Enable pagination/item number information in the Instance's Holding Item list

## Approach

* Since Item list is non-sparse array, we need to make changes in `MLCRenderer` component. [This PR](https://github.com/folio-org/stripes-components/pull/2106) provides correct work both with sparse and non-sparse arrays.
* Next button is disabled when `offset < total` and `total - offset > pageAmount` (number of remaining elements is less than `pageAmount = 200`)

## Refs

[UIIN-2530](https://issues.folio.org/browse/UIIN-2530)

## Screenshots

https://github.com/folio-org/ui-inventory/assets/85172747/be7b1eab-d90c-4cd4-8f9a-586c64422c0f

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,7 @@
 * Fix issue with Relevance sort option. Refs UIIN-2474.
 * Change tenant id to central when opening details of shadow instance. Refs UIIN-2529.
 * "Something went wrong" screen appears when adding Item to Holdings. Fixes UIIN-2534.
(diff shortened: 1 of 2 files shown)


BogdanDenis (approved):

Terala-Priyanka (approved):
It would be nice if the scroll position is restored to the top of the list after retrieving the next set in the list.
Could you check how it works other similar pages?

mariia-aloshyna (approved):

PR #2242, opened 2023-08-22
UIIN-2540: Make the 'enabled' argument always boolean when calling useUserTenantPermissions.

Dmytro-Melnyshyn:
## Purpose
Make the `enabled` argument always boolean when calling `useUserTenantPermissions`.

## Issue
[UIIN-2540](https://issues.folio.org/browse/UIIN-2540)

## Related PRs
https://github.com/folio-org/ui-quick-marc/pull/581
https://github.com/folio-org/stripes-authority-components/pull/100
https://github.com/folio-org/ui-plugin-find-authority/pull/75

### DIFF
--- a/.github/workflows/build-npm-release.yml
+++ b/.github/workflows/build-npm-release.yml
@@ -28,7 +28,7 @@ jobs:
       FOLIO_NPM_REGISTRY: 'https://repository.folio.org/repository/npm-folio/'
       FOLIO_NPM_REGISTRY_AUTH: '//repository.folio.org/repository/npm-folio/'
       FOLIO_MD_REGISTRY: 'https://folio-registry.dev.folio.org'
-      NODEJS_VERSION: '16'
+      NODEJS_VERSION: '18'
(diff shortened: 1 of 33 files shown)


mariia-aloshyna (approved):

mariia-aloshyna:
Hi @Dmytro-Melnyshyn is this fixing the problem with opening instance details for non-consortium users? There is a failed GET request to users/perms right now with okapi header = `undefined`

Dmytro-Melnyshyn:
> Hi @Dmytro-Melnyshyn is this fixing the problem with opening instance details for non-consortium users? There is a failed GET request to users/perms right now with okapi tenant header = undefined

Hi @mariia-aloshyna, this is not tied to that bug. That bug is due to the fact that `_instance.tenantId` doesn't always have `tenantId` in `const instanceTenantId = _instance?.tenantId`, so `useInstanceQuery` doesn't always fire and we don't have instance data.

PR #2252, opened 2023-09-01
UIIN-2560: Display correct number of items in the Instance's Holding Item list

OleksandrHladchenko1:
## Purpose

* Display correct number of items in the Instance's Holding Item list because RMB sends estimated total of items

## Approach

* Set `limit=0` to get exact number of items in `HoldingAccordion` component
* Create separate request to get exact number of items in `ItemsListContainer` component to avoid bugs when interact with pagination

## Refs

[UIIN-2560](https://issues.folio.org/browse/UIIN-2560)

### Before

https://github.com/folio-org/ui-inventory/assets/85172747/5182a162-6a3e-4fdd-ab69-375cd72d9388

### After

https://github.com/folio-org/ui-inventory/assets/85172747/a7ae2362-00b4-4c23-90ff-f7c865968779

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,6 +77,7 @@
 * Adjust behaviour of View source for shared instances. Refs UIIN-2449.
 * Consortial Central Tenant: Handling of Instance Action Menu options. Refs UIIN-2498.
 * ECS: display shadow instances as shared. Refs UIIN-2552.
+* Display correct number of items in the Instance's Holding Item list. Refs UIIN-2560.
 
 ## [9.4.11](https://github.com/folio-org/ui-inventory/tree/v9.4.11) (2023-08-02)
 [Full Changelog](https://github.com/folio-org/ui-inventory/compare/v9.4.10...v9.4.11)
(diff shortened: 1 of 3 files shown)


zburke (approved):
This looks OK to me, but it is important to note that asking for an exact-count via `limit: 0` can be extremely expensive (i.e. slow) when the search corpus is large. From the [RMB docs](folio-org/raml-module-builder/tree/8707dfbbc3c68cd94d8363042cbf33d75a57e4cf#estimated-totalrecords):
> Performance warning: Depending on query and database indexes using `limit = 0` may take long, may put load on the database and should run asynchronously in the front-end. If there isn't a supporting database index it requires a full table scan. In all cases PostgreSQL needs to check the visibility map, see [Slow counting](https://wiki.postgresql.org/wiki/Slow_Counting) and [Index-only_scans](https://wiki.postgresql.org/wiki/Index-only_scans).

Are we confident that an exact count is necessary here? Has this been tested in an environment where individual holdings contain high item-counts, and where the number of items, collectively, is very high? See also [UIIN-2006](https://issues.folio.org/browse/UIIN-2006) for Theo's description of related problems when the holdings-count is high. Do we need to leverage Suspense here to make sure the rest of the UI can render while a large number of slow queries resolves? 

Please note, I _am_ approving this. But I'm also saying, make sure you understand what you're doing here because the impact of this change will be very large for certain datasets that we know exist in production.

PR #2262, opened 2023-09-13
UIIN-2518: Make the Move actions available regardless of how user navigated to the record.

Dmytro-Melnyshyn:
## Purpose
Make the Move actions available regardless of how user navigated to the record.

## Issue
[UIIN-2518](https://issues.folio.org/browse/UIIN-2518)

## Screencast

https://github.com/folio-org/ui-inventory/assets/77053927/1a6e9f98-0aca-46f3-81c0-fd6a3758bd2a

### DIFF
--- a/src/ViewInstance.js
+++ b/src/ViewInstance.js
@@ -460,7 +460,6 @@ class ViewInstance extends React.Component {
       onCopy,
       stripes,
       intl,
-      openedFromBrowse,
       resources: {
         instanceRequests,
       },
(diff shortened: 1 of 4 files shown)


mariia-aloshyna (approved):

src/ViewInstance.js
```diff
@@ -561,31 +561,27 @@ class ViewInstance extends React.Component {
                 disabled={!marcRecord}
               />
             )}
-            {!openedFromBrowse && (
```
BogdanDenis, line 564: I think we can remove all uses of `openedFromBrowse` - one more place is in `showInventoryMenuSection` variable and there rest

BogdanDenis (changes requested):

BogdanDenis (approved):

PR #2266, opened 2023-09-15
UIIN-2525: Consortial Central Tenant: Suppress Item detail view Action menu

mariia-aloshyna:
## Purpose

In the Consortial central tenant, users will be able to view member library item details, but will not be able to act on items unless they change affiliation to a particular member library and have item permissions in that member tenant. To enforce this, we will suppress the Action menu from any Item detail record viewed in the consortial central tenant, no matter what Central tenant Inventory permissions the user has.

## Approach

- Suppress Item details action menu for central tenant
- Add unit tests

## Refs

https://issues.folio.org/browse/UIIN-2525


src/views/ItemView.test.js
```diff
@@ -6,14 +6,22 @@ import { waitFor, screen } from '@folio/jest-config-stripes/testing-library/reac
...
 
 import ItemView from './ItemView';
 
 jest.mock('../Item/ViewItem/ItemAcquisition', () => ({
   ItemAcquisition: jest.fn(() => 'ItemAcquisition'),
 }));
+jest.mock('@folio/stripes/core', () => ({
+  ...jest.requireActual('@folio/stripes/core'),
+  checkIfUserInCentralTenant: jest.fn(),
+}));
```
Dmytro-Melnyshyn, line 24: Please don't override the existing `stripes/core` mock, you can remove `<StripesContext.Provider value={stripesStub}>` from this file and mock `checkIfUserInCentralTenant` in `stripesCore.mock.js`. So this file will only contain one version of `stripes` mock.

OleksandrHladchenko1 (approved):

src/views/ItemView.test.js
mariia-aloshyna: good point, fixed

mariia-aloshyna:

Dmytro-Melnyshyn (approved):

PR #2267, opened 2023-09-18
UIIN-2572: Consortial Central Tenant: Hide Instance Action Menu New Request option

mariia-aloshyna:
## Purpose

To suppress `New request` action item from Consortial Central Tenant

## Approach

- Check if user is in central tenant and hide the action item when true

## Refs

https://issues.folio.org/browse/UIIN-2572


OleksandrHladchenko1 (approved):

mariia-aloshyna:

PR #2274, opened 2023-09-22
UIIN-2570: Disable View source button for Source = FOLIO instances

OleksandrHladchenko1:
## Purpose

* Disable "View source" button for Source = FOLIO instances

## Approach

* Changed prop name to `isDisabled` which is correct prop for `ActionItem` component

## Refs

[UIIN-2570](https://issues.folio.org/browse/UIIN-2570)

## Screenshots

https://github.com/folio-org/ui-inventory/assets/85172747/4c582b39-492f-404c-91f1-4b1b66e5d17c

### DIFF
--- a/src/ViewInstance.js
+++ b/src/ViewInstance.js
@@ -603,7 +603,7 @@ class ViewInstance extends React.Component {
                   onToggle();
                   this.handleViewSource(e, instance);
                 }}
-                disabled={!marcRecord}
+                isDisabled={!marcRecord}
               />
             )}
             {canMoveItems && (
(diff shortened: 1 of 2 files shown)


UladzislauKutarkin (approved):

src/ViewInstance.js
```diff
@@ -603,7 +603,7 @@ class ViewInstance extends React.Component {
                   onToggle();
                   this.handleViewSource(e, instance);
                 }}
-                disabled={!marcRecord}
+                isDisabled={!marcRecord}
```
mariia-aloshyna, line 606: Please see the bug description's NOTE, it's better to rename the component's prop to just `disabled`, some action items also pass `disabled` instead of `isDisabled`

mariia-aloshyna (approved):

PR #2283, opened 2023-10-03
UIIN-2407 Use different call number types in facets

BogdanDenis:
## Description
Change `callNumberType` parameter for facets when browsing by different call number types

## Screenshots

https://github.com/folio-org/ui-inventory/assets/19309423/3dc6a1ea-09f8-40ac-8f9e-5c751c0ecb4a

## Issues
[UIIN-2407](https://issues.folio.org/browse/UIIN-2407)

### DIFF
--- a/src/withFacets.js
+++ b/src/withFacets.js
@@ -149,7 +149,8 @@ function withFacets(WrappedComponent) {
       } else if ([FACETS.SUBJECTS_SHARED, FACETS.SUBJECTS_HELD_BY].includes(facetName)) {
         params.query = 'value=*';
       } else if (cqlQuery && Object.values(browseCallNumberOptions).includes(queryIndex)) {
-        params.query = 'callNumber=""';
+        const callNumberType = queryIndex === browseCallNumberOptions.CALL_NUMBERS ? '' : queryIndex;
+        params.query = `callNumberType="${callNumberType}"`;
       } else if (cqlQuery && queryIndex !== browseModeOptions.CALL_NUMBERS) {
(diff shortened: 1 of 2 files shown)


src/withFacets.js
```diff
@@ -149,7 +149,8 @@ function withFacets(WrappedComponent) {
       } else if ([FACETS.SUBJECTS_SHARED, FACETS.SUBJECTS_HELD_BY].includes(facetName)) {
         params.query = 'value=*';
       } else if (cqlQuery && Object.values(browseCallNumberOptions).includes(queryIndex)) {
-        params.query = 'callNumber=""';
+        const callNumberType = queryIndex === browseCallNumberOptions.CALL_NUMBERS ? '' : queryIndex;
+        params.query = `callNumberType="${callNumberType}"`;
```
Dmytro-Melnyshyn, line 153: Should the `query` also include the selected facet options as implemented in the `Search` lockup?
e.g. `query: callNumberType="local" and shared=="true"`

BogdanDenis: Possibly, but that's out of scope of this ticket. This needs to additionally investigated and estimated

Dmytro-Melnyshyn: I mean only the `Shared` facet. If we only use the `Shared` facet in the `Search` lookup and select `Yes`, the facet request query will contain the selected options, but the `Shared` facet in the `Browse` search doesn't append selected options to facet request.

Dmytro-Melnyshyn (approved):

PR #2284, opened 2023-10-03
UIIN-2452: Enable/disable consortial holdings/item actions based on User permissions

OleksandrHladchenko1:
## Purpose

* In a consortium environment, Instance records may contain holdings from multiple institutions and users from one library may need to evaluate materials that are held by different institutions.
* Users in a central consortial environment should be able to view holdings and items for all member libraries, and depending on permissions, also create and edit holdings and items.
* Users in a local member library should be able to view their own holdings and items first, followed by other consortial member library holdings and items.

## Approach

* Added `tenantId` parameter to `useBoundWithHoldings` to dynamically change **X-Okapi-Tenant** header
* In `ViewInstance` component we fetch all the assigned permissions of all user's tenants using `getUserTenantsPermissions` function
* In `getCurrentTenantPermissions` function we get all the assigned permissions of all user's tenants if `isUserInConsortiumModed === true` otherwise we get just all user's permissions
* Added `isViewHoldingsDisabled` and `isAddItemDisabled` props to `HoldingButtonsGroup` component to disable "View Holdings" and "Add item" buttons when member tenant doesn't have permissions
* Added `isBarcodeAsHotlink` prop to `ItemBarcode` component to display barcode as a hotlink or not depends on user's permission
* Added `disabled` prop to `InstanceNewHolding` component to disable "Add holding" button if user doesn't have permissions
* In order to update affiliation correctly, replaced `Link` component with `Button` and added styles so it behaves as a hotlink 
* When user wants to view or create Item/Holding we call `updateAffiliation` function to switch affiliations if user in on the other member tenant.
* Updated and added unit tests

## Refs

[UIIN-2452](https://issues.folio.org/browse/UIIN-2452)

## Screenshots

https://github.com/folio-org/ui-inventory/assets/85172747/371803b6-3861-4384-8bfc-89f56e324d0a


OleksandrHladchenko1:
Hi @BogdanDenis @Dmytro-Melnyshyn Please take a look at this PR

src/Holding/CreateHolding/CreateHolding.js
```diff
@@ -19,26 +19,28 @@ import {
...
   const { instance, isLoading: isInstanceLoading } = useInstance(instanceId);
   const sourceId = referenceData.holdingsSourcesByName?.FOLIO?.id;
 
-  const onCancel = useCallback(() => {
-    history.push({
-      pathname: `/inventory/view/${instanceId}`,
-      search: location.search,
-    });
+  const goBack = useCallback(() => {
+    window.location.href = `/inventory/view/${instanceId}${location.search}`;
```
BogdanDenis, line 37: Why do you use `window.location.href` instead of `react-router` API?

BogdanDenis (changes requested):
Please add an entry to CHANGELOG
Also, there's an unhandled exception when closing a Holding:
Tested on https://folio-dev-spitfire-2nd-consortium.ci.folio.org/
https://github.com/folio-org/ui-inventory/assets/19309423/4cf6e978-a331-46eb-ac24-781128a277e0

src/Instance/HoldingsList/Holding/HoldingContainer.js
```diff
@@ -102,18 +112,12 @@ const HoldingContainer = ({
   ...rest
 }) => {
   const onViewHolding = useCallback(() => {
-    history.push({
-      pathname: `/inventory/view/${instance.id}/${holding.id}`,
-      search: location.search,
-    });
+    window.location.href = `/inventory/view/${instance.id}/${holding.id}${location.search}`;
```
Dmytro-Melnyshyn, line 115: this reloads the page

src/Instance/HoldingsList/Holding/HoldingButtonsGroup.js
```diff
@@ -42,7 +47,8 @@ const HoldingButtonsGroup = ({
       <Button
         id={`clickable-view-holdings-${holding.id}`}
         data-test-view-holdings
-        onClick={onViewHolding}
+        onClick={() => updateAffiliation(stripes.okapi, tenantId, onViewHolding)}
+        disabled={isViewHoldingsDisabled}
```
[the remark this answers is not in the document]
zburke: > I think we shouldn't display elements if a user doesn't have permission. I guess it's a FOLIO pattern.

Yep, that's the FOLIO UX I'm familiar with. I will follow up with the PO here as I see the story was written specifically with disabled elements, and I've seen similar stories in ui-users. We need to figure out if we are purposely abandoning that pattern or if this was just a goof.

src/Instance/HoldingsList/Holding/HoldingContainer.js
zburke: Good catch, @Dmytro-Melnyshyn!

mariia-aloshyna: You are right, @Dmytro-Melnyshyn, this is what @OleksandrHladchenko1 supposed to do when used `window.location.href` instead of `history.push`. The page needs to be refreshed because user's affiliation changes, tenant is updated, new permissions are fetched so it is similar to switching affiliation as it is done in `ui-consortia-settings`

src/Holding/CreateHolding/CreateHolding.js
mariia-aloshyna: @BogdanDenis To refresh the page, see the comment from Dmytro please, I described the process there

src/Instance/HoldingsList/Holding/HoldingButtonsGroup.js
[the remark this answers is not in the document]
mariia-aloshyna: @zburke and @Dmytro-Melnyshyn This is a requirement for this story, but we discussed with PO that suppressing these buttons from a user instead of disabling them will make the page kinda less noisy and will free up more space. We decided to go with the disabled buttons in the main release and after further testing and feedback from users, we might end up hiding these buttons.

src/Instance/InstanceDetails/InstanceNewHolding/InstanceNewHolding.js
```diff
@@ -1,38 +1,46 @@
...
   instance,
+  tenantId,
+  disabled,
 }) => {
   const intl = useIntl();
+  const stripes = useStripes();
 
   const label = intl.formatMessage({ id: 'ui-inventory.addHoldings' });
 
+  const onNewHolding = useCallback(() => {
```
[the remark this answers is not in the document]
mariia-aloshyna: @Dmytro-Melnyshyn it is used on line 42

mariia-aloshyna:
> Please add an entry to CHANGELOG Also, there's an unhandled exception when closing a Holding: Tested on https://folio-dev-spitfire-2nd-consortium.ci.folio.org/
> 
>  chrome_nBkpAkTWAP.mp4

fixed

src/Instance/HoldingsList/Holding/HoldingContainer.js
mariia-aloshyna: Okay, my comment just vanished for some reason.
So the page is supposed to be reloaded because the affiliation is changed, new tenant permissions are fetched, and as a result, other modules are available for the user. The same logic is implemented in `ui-consortia-settings` module when switching affiliation.

mariia-aloshyna:
Hi @Dmytro-Melnyshyn and @BogdanDenis I fixed your comments, please, double-check

zburke (commented):
I cannot separate the new feature work in this PR from the new UX patterns it introduces (disabled elements instead of hidden elements; deliberately reloading the page). There is no committee/policy/whatever for reviewing new UX patterns, but it is my strong feeling that they should not just be slipped in like this. 

Sorry, @mariia-aloshyna, I know you're just trying to implement this ticket as-specified. I won't stand in the way of it, but I also cannot approve it. I will follow up with the PO.

BogdanDenis (approved):
Agree with @zburke that potentially `validateUser` could be used to re-load modules without page reload.
For this though some modifications need to be made to that function, because `validateUser` takes `tenant` from current session
https://github.com/folio-org/stripes-core/blob/master/src/loginServices.js#L411

If we're going to continue with page reload - please create a tech debt story and add TODO comments where that reload happens to explain why the reload is needed and link the tech debt story

zburke:
> For this though some modifications need to be made to that function, because `validateUser` takes `tenant` from current session

That's true, @BogdanDenis, but that feels like a bug: I expected `validateUser()` to use the tenant it is given and only pull from the session if no argument is specified, similar to the `useOkapiKy` implementation in #1348. In fact it's the reverse, using the value pulled from the session by default and only accepting the argument if the session's value is empty. That's puzzling given that `createOkapiSession` always stores a value for `tenant` in the session, so I can't come up with any circumstance when `validateUser()` would use its argument.

@NikitaSedyx, can you provide some context for why it was written this way since it seems deliberate? It would be nice to reuse `validateUser()` if possible, but if we have different needs (sometimes we want the value from storage to be the default, sometimes we want a specific value to be the default) then we can copy-and-tweak the existing function.

src/Holding/CreateHolding/CreateHolding.js
```diff
@@ -32,13 +31,19 @@ const CreateHolding = ({
...
     history.push({
       pathname: `/inventory/view/${instanceId}`,
       search: location.search,
     });
   }, [location.search, instanceId]);
 
+  const onCancel = useCallback(() => {
+    const { location: { state: { tenantFrom } } } = history;
+
+    switchAffiliation(stripes, tenantFrom, goBack).then();
```
Dmytro-Melnyshyn, line 44: Is `then` still needed?

src/Instance/InstanceDetails/InstanceDetails.js
```diff
@@ -195,11 +200,18 @@ const InstanceDetails = forwardRef(({
...
+              <InstanceNewHolding
+                instance={instance}
+                disabled={!canCreateHoldings}
+                tenantId={tenantId}
+              />
             )}
 
             {isConsortialHoldingsVisible && (
-              <ConsortialHoldings instance={instance} />
+               <ConsortialHoldings
```
Dmytro-Melnyshyn, line 211: Please fix eslint warnings.

src/Item/CreateItem/CreateItem.js
```diff
@@ -40,12 +44,16 @@ const CreateItem = ({
...
+    push({
       pathname: `/inventory/view/${instanceId}`,
-      search: location.search,
+      search,
     });
-  }, [location.search, instanceId]);
+  }, [instanceId, search]);
+
+  const onCancel = useCallback(() => {
+    switchAffiliation(stripes, tenantFrom, goBack).then();
```
Dmytro-Melnyshyn, line 55: Is `then` still needed?

src/routes/ItemRoute.js
```diff
@@ -155,19 +158,26 @@ class ItemRoute extends React.Component {
...
+  }
+
+  onClose = () => {
+    const {
+      stripes,
+      location: { state: { tenantFrom } },
+    } = this.props;
 
-    goTo(`${path}${search}`);
+    switchAffiliation(stripes, tenantFrom, this.goBack).then();
```
Dmytro-Melnyshyn, line 180: then

src/utils.js
```diff
@@ -21,6 +21,8 @@ import {
 } from 'lodash';
 import moment from 'moment';
 
+import { updateTenant } from '@folio/stripes/core';
+import { validateUser } from '@folio/stripes-core/src/loginServices';
```
Dmytro-Melnyshyn, line 25: Let's import `validateUser` from `'@folio/stripes/core';`

src/Instance/HoldingsList/HoldingsListContainer.test.js
```diff
@@ -39,7 +44,10 @@ describe('HoldingsListContainer', () => {
 
     render(
       <MemoryRouter>
-        <HoldingsListContainer instance={instance} />
+        <HoldingsListContainer
+          instance={instance}
+          userTenantPermissions={userTenantPermissions}
```
mariia-aloshyna, line 49: Is `userTenantPermissions` still used in `HoldingsListContainer`?

mariia-aloshyna (approved):

Dmytro-Melnyshyn (approved):

src/utils.js
```diff
@@ -21,6 +21,10 @@ import {
 } from 'lodash';
 import moment from 'moment';
 
+import {
+  updateTenant,
+  validateUser,
```
BogdanDenis, line 26: I think `validateUser` is not currently exported from `stripes-core` so we need a PR there to export it

BogdanDenis (changes requested):

src/utils.js
OleksandrHladchenko1: Hi @BogdanDenis here it is
https://github.com/folio-org/stripes-core/pull/1351

BogdanDenis (approved):

PR #2287, opened 2023-10-04
UIIN-2395 Make Call numbers browse Held By facet filterable

BogdanDenis:
## Description
Make Call numbers browse Held By facet filterable

## Issues
[UIIN-2395](https://issues.folio.org/browse/UIIN-2395)


Dmytro-Melnyshyn:
We can close this PR, all the necessary changes have been made.

PR #2293, opened 2023-10-05
UIIN-2592 Counter in the "Shared" facet doesn't update during switching between different browse options

BogdanDenis:
## Description
Counter in the "Shared" facet doesn't update during switching between different browse options

## Approach
Reset accordion state when it is unmounted. More info in https://github.com/folio-org/stripes-components/pull/2138
Save last search params for facet when that facet is fetched. This was done incorrectly previously

## Screenshots

https://github.com/folio-org/ui-inventory/assets/19309423/8551e8e6-83d0-4625-9240-2358103e3811

## Issues
[UIIN-2592](https://issues.folio.org/browse/UIIN-2592)

## Related PRs
https://github.com/folio-org/stripes-components/pull/2138


mariia-aloshyna (approved):

src/common/hooks/useFacets.js
```diff
@@ -180,8 +183,7 @@ const useFacets = (
       !facetToOpen.match(/updatedDate/i)
     ) {
       handleFetchFacets({ facetToOpen });
-    } else {
-      prevUrl.current[facetToClose] = location.search;
+      prevUrl[facetToOpen] = location.search;
```
Dmytro-Melnyshyn, line 186:
```suggestion
      prevUrl.current[facetToOpen] = location.search;
```

Dmytro-Melnyshyn (approved):

PR #2301, opened 2023-10-11
UIIN-2775: Add confirmation modal when local instance has linked MARC Authorities during the sharing

OleksandrHladchenko1:
## Purpose

* In a consortium environment, fields in local MARC bib records may be linked to local MARC authority records. However, Instances with source = MARC may also be promoted to be shared to the rest of the consortium. When the Instance with source = MARC is shared, we need to unlink the controlled fields in the MARC bibliographic record from the local MARC authority records and show a confirmation modal in the UI.

## Approach

* Added `AUTHORITY_LINKED_FIELDS` array that contains all the fields which can be linked to MARC Authorities
* `findLinkedAuthorities` return total amount of linked fields
* If there is no linked MARC Authorities - just continue sharing
* If the local instance has linked MARC Authorities when user confirm sharing, show confirmation modal
* When proceed unlinking, then start sharing the instance

## Refs

[UIIN-2575](https://issues.folio.org/browse/UIIN-2575)

## Screenshots

https://github.com/folio-org/ui-inventory/assets/85172747/462f923b-87f1-4619-ad6b-e4aadb4c8a62


BogdanDenis (approved):
Please also add an entry to changelog

Dmytro-Melnyshyn (approved):

mariia-aloshyna (approved):

PR #2305, opened 2023-10-13
UIIN-2453:Instance 3rd pane: Adjust behavior when returning to instance from holdings/item full screen

mariia-aloshyna:
## Purpose

In a consortium environment, Instance records may contain holdings from multiple institutions and users from one library may need to evaluate materials that are held by different institutions. Users affiliated with various member libraries should be able to view, create, and edit holdings for other member libraries, depending on their permissions for that affiliation, and then return to the Instance view that they started from.

## Approach

- Use `sessionStorage` to store state of holdings accordions for a particular instance so a user could get back to the view he started from when navigated away from instance details view
- Create a custom hook to set and get holdings accordions state to the `sessionStorage`
- When the instance id changes - reset the consortial holdings accordion state

## Refs

https://issues.folio.org/browse/UIIN-2453

## Screenshots

https://github.com/folio-org/ui-inventory/assets/55138456/e353deca-bece-4b07-9d4b-8d4093b90834


OleksandrHladchenko1 (approved):

src/ViewInstance.js
```diff
@@ -900,6 +900,7 @@ class ViewInstance extends React.Component {
       },
     ];
     const isInstanceLoading = this.state.isLoading || !instance || isCentralTenantPermissionsLoading;
+    const keyInStorageToHoldingsAccsState = ['holdings'];
```
BogdanDenis, line 903: It looks like it could be a constant which can be imported and used directly in `<HoldingAccordion>` without passing it through the component tree. Is there any reason to have it passed as a prop?

PR #2310, opened 2023-10-17
UIIN-2531 Reset CheckboxFacet state.more when user resets search form and fewer facet options are loaded.

BogdanDenis:
## Description
Fix issue with "+ More" button disappearing after user selects some Effective location filters and resets search form.

## Approach
When user clicks "+ More" button `state.more` is calculated to increase by 5 every click. When user loads more `state.more` is now `dataOptions.length + 5`. And when user clicks "Reset all" - 5 initial options are rendered, but `state.more` is still more than 5 so `showMore={filteredOptions.length > more}` condition fails.
So the solution is to reset `state.more` to default value when component is re-rendered with fewer options

## Screnshots

https://github.com/folio-org/ui-inventory/assets/19309423/e77d6772-8c85-413d-bdc1-24c86ac6fd93

## Issues
[UIIN-2531](https://issues.folio.org/browse/UIIN-2531)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change history for ui-inventory
 
+## [10.1.0] (IN PROGRESS)
+
+* Reset CheckboxFacet state.more when user resets search form and fewer facet options are loaded. Fixes UIIN-2531.
+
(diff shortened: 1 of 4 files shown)


Dmytro-Melnyshyn (approved):

zburke (approved):
LGTM. Thank you for describing the bug and how the fix works in the PR and in the comments!

PR #2311, opened 2023-10-17
UIIN-2600: Remove error message after switch from Instance Edit screen to another app

OleksandrHladchenko1:
## Purpose

* Remove error message "Something went wrong" after switch from Instance Edit screen to another app

## Approach

* Added optional chaining for instance form values

## Refs

[UIIN-2600](https://issues.folio.org/browse/UIIN-2600)

## Screenshots

https://github.com/folio-org/ui-inventory/assets/85172747/19f09e3e-a38b-46a2-a88b-b33bb1e78912


src/edit/InstanceForm.js
```diff
@@ -309,31 +309,31 @@ class InstanceForm extends React.Component {
       return ref || {};
     };
 
-    const instanceTypeOptions = referenceTables.instanceTypes ? referenceTables.instanceTypes.map(
+    const instanceTypeOptions = referenceTables.instanceTypes ? referenceTables.instanceTypes?.map(
```
zburke, line 312: Pick one: we don't need both a ternary and optional chaining here.

src/edit/InstanceForm.js
```diff
@@ -342,7 +342,7 @@ class InstanceForm extends React.Component {
     ) : [];
 
     const statisticalCodeOptions = referenceTables.statisticalCodes
-      .map(
+      ?.map(
```
zburke, line 345: Ah, now I see what happened above: here, there was no ternary and you solved the bug with optional chaining, and so added optional chaining everywhere. It doesn't matter which we choose, but let's pick one pattern and stick with it.

zburke (changes requested):
We went from not enough null-checking to too much. Let's pick either ternary or optional chaining. It doesn't matter which but we don't need both.

mariia-aloshyna (approved):

BogdanDenis (approved):

PR #2319, opened 2023-10-24
UIIN-2452: Fix creating/editing/duplicating holdings/items for different member tenants (follow-up)

OleksandrHladchenko1:
## Approach

When user completes duplicating/editing or creating holdings/items, switch to the previous affiliation


mariia-aloshyna (approved):

src/Instance/InstanceDetails/MemberTenantHoldings/MemberTenantHoldings.js
```diff
@@ -40,10 +40,10 @@ const MemberTenantHoldings = ({
   const { holdingsRecords, isLoading } = useInstanceHoldingsQuery(instance?.id, { tenantId: id });
   const isUserInCentralTenant = checkIfUserInCentralTenant(stripes);
 
-  const canViewHoldings = hasMemberTenantPermission('ui-inventory.instance.view', id, userTenantPermissions);
-  const canCreateItem = hasMemberTenantPermission('ui-inventory.item.edit', id, userTenantPermissions);
-  const canCreateHoldings = hasMemberTenantPermission('ui-inventory.holdings.edit', id, userTenantPermissions);
-  const canViewItems = hasMemberTenantPermission('ui-inventory.instance.view', id, userTenantPermissions);
+  const canViewHoldings = hasMemberTenantPermission('ui-inventory.holdings.create', id, userTenantPermissions);
```
BogdanDenis, line 43: Please also update variable names

BogdanDenis (changes requested):

src/Instance/InstanceDetails/MemberTenantHoldings/MemberTenantHoldings.js
OleksandrHladchenko1: Hi @BogdanDenis To view and create holdings we need "View, create holdings" permission (ui-inventory.holdings.create).  **canViewHoldings** is used for "View holdings" button. **canCreateHoldings** is used for "Add holdings" button

UladzislauKutarkin (approved):

src/Instance/InstanceDetails/MemberTenantHoldings/MemberTenantHoldings.js
BogdanDenis: It's confusing to me why there are two variables that are checking for the same permission and they are called differently.
Additionally, from variable name `canViewHoldings` I expect it to be true only when user has view permissions, but not edit or create.

mariia-aloshyna: @BogdanDenis I was also confused about the fact that to view holdings you actually need to check for View and **create** permission, we discussed it with PO and figured out that there is no separate permission to just view holdings. I suggest renaming this variable to `canViewAndCreateHoldings` and using it in two places

BogdanDenis: @OleksandrHladchenko1 please review comment from @mariia-aloshyna

OleksandrHladchenko1:
@BogdanDenis please, review one more time

BogdanDenis (approved):

PR #2323, opened 2023-10-27
UIIN-2634: Set tenant from `stripes.okapi.tenant` if `location.state` is empty (follow-up)

OleksandrHladchenko1:
## Approach

* Move manifest from `ItemRoute` to `ItemView` component in order to pass `!{tenantTo}` value
* Added `tenantTo` prop to `ViewHoldingsRecord` component to simplify reading `tenant` field in manifest

## Refs

[UIIN-2634](https://issues.folio.org/browse/UIIN-2634)


src/ViewHoldingsRecord.js
```diff
@@ -81,7 +81,14 @@ class ViewHoldingsRecord extends React.Component {
...
       resourceShouldRefresh: false,
       accumulate: true,
-      tenant: '!{location.state.tenantTo}'
+      tenant: (_q, _p, _r, _l, props) => {
+        const {
+          stripes,
+          location,
+        } = props;
+
+        return location?.state?.tenantTo || stripes.okapi.tenant;
```
mariia-aloshyna, line 90: `stripes-connect` already does this check
https://github.com/folio-org/stripes-connect/blob/master/OkapiResource.js#L48

mariia-aloshyna (changes requested):

vashjs (approved):
Please update Changelog.md

OleksandrHladchenko1:
> Please update Changelog.md

It's follow-up change. CHANGELOG is already updated

src/views/ItemView.js
```diff
@@ -86,11 +90,154 @@ import {
   WarningMessage,
   AdministrativeNoteList,
   ItemViewSubheader,
+  PaneLoading,
 } from '../components';
 
+const getRequestsPath = `circulation/requests?query=(itemId==:{itemid}) and status==(${requestsStatusString}) sortby requestDate desc&limit=1`;
```
mariia-aloshyna, line 96: I understand that this is just code replaced from another component, but I don't see any reason to have a separate variable for this path, it's used in one place only and the name of the variable looks like this is a function rather than a constant. I suggest either putting it to manifest or renaming it.

mariia-aloshyna (approved):

PR #2324, opened 2023-10-30
UIIN-2486 Added support for `containsAny` match option in Advanced search

BogdanDenis:


vashjs (approved):

UladzislauKutarkin (approved):

src/constants.js
```diff
@@ -466,146 +466,175 @@ export const fieldSearchConfigurations = {
     exactPhrase: 'keyword=="%{query.query}"',
     containsAll: 'keyword all "%{query.query}"',
     startsWith: 'keyword all "%{query.query}*"',
+    containsAny: 'keyword any "%{query.query}"',
   },
   contributor: {
     exactPhrase: 'contributors.name=="%{query.query}"',
     containsAll: 'contributors.name="*%{query.query}*"',
     startsWith: 'contributors.name="%{query.query}*"',
+    containsAny: 'contributors.name any "*%{query.query}*"',
```
Dmytro-Melnyshyn, line 475: [contributors.name](https://github.com/folio-org/mod-search#instance-search-options) is a `term` field, `any` operator can be used only for a [full text](https://github.com/folio-org/mod-search#cql-query-operators) field.

mariia-aloshyna (approved):

src/constants.js
BogdanDenis: Discussed it with Christine and she said that we should perform the search anyways even if there are no results

Dmytro-Melnyshyn (approved):

PR #2328, opened 2023-11-03
UIIN-2590: ECS: Show info message when user in member tenant tries to view shared instance details without permission

mariia-aloshyna:
## Purpose

If a user is on a member tenant and wants to view details of a shared instance retrieved in an Inventory search, the user must have the central tenant permission: Inventory: Can view instances, holdings, items. Otherwise they see 3 dots. This bugfix replaces the 3 dots with a better indication of the problem.

## Approach

- Implement `InstanceWarningPane` and `InstanceLoadingPane` as separate components
- `InstanceWarningPane` shows the pure instance pane with a warning banner instead of details
- `InstanceLoadingPane` shows the pure instance pane with a loading spinner instead of details
- Move the check of instance loading or error upper to `ViewInstanceWrapper` to avoid redundant instance properties checks when the instance is empty yet.
- Check whether the user lacks permission to view shared instances and display a warning message if so. I wanted to check it using `centralTenantPermissions` but realized that this request also can be failed, so decided to just check for response error status code.
- Adjust unit tests

## Refs

https://issues.folio.org/browse/UIIN-2590

## Screenshots

image


src/ViewInstanceWrapper.js
```diff
@@ -17,7 +17,7 @@ const ViewInstanceWrapper = (props) => {
   const userId = stripes?.user?.user?.id;
   const centralTenantId = stripes.user.user?.consortium?.centralTenantId;
   const consortiumId = stripes.user.user?.consortium?.id;
-  const { instance, isLoading, refetch } = useInstance(id);
+  const { instance, isLoading, refetch, isError, error } = useInstance(id);
```
OleksandrHladchenko1, line 20:
```suggestion
  const {
    instance,
    isLoading,
    refetch,
    isError,
    error,
  } = useInstance(id);
```

src/common/hooks/useInstance.js
```diff
@@ -9,7 +9,7 @@ const useInstance = (id) => {
   const instanceTenantId = _instance?.tenantId;
   const isShared = _instance?.shared;
 
-  const { isLoading: isInstanceLoading, instance: data, refetch } = useInstanceQuery(
+  const { isLoading: isInstanceLoading, instance: data, refetch, ...rest } = useInstanceQuery(
```
OleksandrHladchenko1, line 12:
```suggestion
  const {
    isLoading: isInstanceLoading,
    instance: data,
    refetch,
    ...rest,
  } = useInstanceQuery(
```

src/common/hooks/useInstanceQuery/useInstanceQuery.js
```diff
@@ -8,7 +8,7 @@ const useInstanceQuery = (instanceId, { tenantId = '' } = {}, options = {}) => {
   const ky = useTenantKy({ tenantId });
   const [namespace] = useNamespace({ key: 'instance' });
 
-  const { isLoading, data: instance = {}, refetch } = useQuery(
+  const { isLoading, data: instance = {}, refetch, ...rest } = useQuery(
```
OleksandrHladchenko1, line 11: Here the same

BogdanDenis (approved):

OleksandrHladchenko1 (approved):

PR #2337, opened 2023-11-07
UIIN-2647: Share instance when there are no linked MARC Authorities (follow-up)

OleksandrHladchenko1:
## Approach

* If there are no linked MARC Authorities => proceed with sharing.
* If there is at least one local linked MARC Authority => show another modal to proceed sharing with unlinking MARC Authorities.
* If there are only shared linked MARC Authorities => proceed with sharing.

## Refs

[UIIN-2647](https://issues.folio.org/browse/UIIN-2647)


mariia-aloshyna (approved):

src/ViewInstance.js
```diff
@@ -527,26 +527,31 @@ class ViewInstance extends React.Component {
...
-        });
-      } else {
-        this.handleShareLocalInstance(instance);
-      }
-    });
+  checkIfHasLinkedAuthorities = () => {
+    const { selectedInstance } = this.props;
+    const authorityIds = getLinkedAuthorityIds(selectedInstance);
+
+    if (!isEmpty(authorityIds)) {
```
BogdanDenis, line 534: Let's reduce the depth a little:

```
if (isEmpty(authorityIds)) {
  return this.handleShareLocalInstance(selectedInstance);
}

this.props.mutator.authorities.GET({
        params: {
          query: `id==(${authorityIds.join(' or ')})`,
        }
      }).then(({ authorities }) => {
        const localAuthorities = authorities.filter(authority => !authority.source.startsWith(CONSORTIUM_PREFIX));

        if (localAuthorities.length) {
          this.setState({
            linkedAuthoritiesLength: localAuthorities.length,
            isShareLocalInstanceModalOpen: false,
            isUnlinkAuthoritiesModalOpen: true,
          });
        } else {
          this.handleShareLocalInstance(selectedInstance);
        }
      });
    }
```

BogdanDenis (changes requested):

BogdanDenis (approved):

PR #2340, opened 2023-11-09
UIIN-2680: Fetch sharing of local instance status in the context of member tenant

mariia-aloshyna:
## Purpose

A user gets 403 error when trying to fetch a local instance sharing status. Send requests in the context of member tenant instead of central tenant.

## Approach

- Send requests in the context of member tenant instead of central tenant.

## Refs

https://issues.folio.org/browse/UIIN-2680

### DIFF
--- a/src/ViewInstance.js
+++ b/src/ViewInstance.js
@@ -45,7 +45,6 @@ import {
   indentifierTypeNames,
   INSTANCE_SHARING_STATUSES,
   layers,
-  OKAPI_TENANT_HEADER,
   REQUEST_OPEN_STATUSES,
 } from './constants';
 import { DataContext } from './contexts';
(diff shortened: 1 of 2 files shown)


BogdanDenis (approved):

src/ViewInstance.js
```diff
@@ -447,11 +447,6 @@ class ViewInstance extends React.Component {
   checkInstanceSharingProgress = ({ sourceTenantId, instanceIdentifier }) => {
     return this.props.mutator.shareInstance.GET({
       params: { sourceTenantId, instanceIdentifier },
-      headers: {
-        [OKAPI_TENANT_HEADER]: this.props.centralTenantId,
```
Dmytro-Melnyshyn, line 451: `OKAPI_TENANT_HEADER` is never used in imports.

Dmytro-Melnyshyn (approved):

PR #2331, opened 2023-11-06
UIIN-2497 After creating a fast add instance record, only an empty detail view is displayed (third column)

BogdanDenis:
## Description
 Show Instance record after creating with Fast add option

## Screenshots

https://github.com/folio-org/ui-inventory/assets/19309423/393659cb-ca08-412d-a853-d6a5cb89fadf

## Issues
[UIIN-2497](https://issues.folio.org/browse/UIIN-2497)

PR #2344, opened 2023-11-10
UIIN-2497 Fix exception when closing Fast Add record modal

BogdanDenis:
## Description
When closing Fast Add modal, function `toggleNewFastAddModal` argument is undefined

## Issues
[UIIN-2497](https://issues.folio.org/browse/UIIN-2497)


src/components/InstancesList/InstancesList.js
```diff
@@ -548,7 +548,7 @@ class InstancesList extends React.Component {
     }
   }
 
-  toggleNewFastAddModal = ({ instanceRecord }) => {
+  toggleNewFastAddModal = ({ instanceRecord } = {}) => {
```
mariia-aloshyna, line 551: If I'm not mistaken on line 739 `buildOnClickHandler` calls this function with arg `this.context.sendCallout` which is a function, not an object with `instanceRecord` prop. Looks like something is not correct here

BogdanDenis: Nice catch @mariia-aloshyna, I added a separate function to handle closing of the modal. It will always receive the same type of argument

mariia-aloshyna:
Update CHANGELOG also please

BogdanDenis:
> Update CHANGELOG also please

This is a follow-up request to this one https://github.com/folio-org/ui-inventory/pull/2331

Terala-Priyanka (changes requested):
Please update changelog.md file

mariia-aloshyna:
> Please update changelog.md file

Hi @Terala-Priyanka see a comment from Denys above, it's follow-up PR, changelog has been already updated

Terala-Priyanka (approved):

PR #2350, opened 2023-11-15
UIIN-2604 Ignored hot key command on edit fields.

przemyslawturek:
Ignored hot key command  (ALT + n) on edit fields.

## Purpose
Prevent execute hot key command on edit fields to able input special characters like polish "ń" 

## Approach
By detect target event is edit field 

## Refs
[UIIN-2604](https://issues.folio.org/browse/UIIN-2604)

## Screenshots

Inventory_polish_special_charakter_alt_n

### DIFF
--- a/src/components/InstancesList/InstancesList.js
+++ b/src/components/InstancesList/InstancesList.js
@@ -56,7 +56,6 @@ import {
   marshalInstance,
   omitFromArray,
   isTestEnv,
-  handleKeyCommand,
   buildSingleItemQuery,
   isUserInConsortiumMode,
 } from '../../utils';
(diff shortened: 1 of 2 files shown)


src/components/InstancesList/InstancesList.js
```diff
@@ -1174,11 +1173,15 @@ class InstancesList extends React.Component {
     const shortcuts = [
       {
         name: 'new',
-        handler: handleKeyCommand(() => {
+        handler: (e) => {
+          const nodeName = e.target?.nodeName || '';
+          const editTextField = (nodeName === 'TEXTAREA') || (nodeName === 'INPUT');
```
BogdanDenis, line 1178: Let's rename it to `isEditTextField` to be consistent with boolean flag naming

BogdanDenis (changes requested):

src/components/InstancesList/InstancesList.js
przemyslawturek: Yeah sure

BogdanDenis (approved):

mariia-aloshyna (commented):
What do you think about passing to `HasCommand` in `InstancesList` component custom `checkScope` func to check whether the active element is not input or textarea? So it will work not only for `n` letter but for others as well

   ```
 const checkScope = () => {
      const ignoreElements = ['TEXTAREA', 'INPUT'];

      return !ignoreElements.includes(document.activeElement.tagName);
    };
```

przemyslawturek:
I think it couse to the same results, but checking scope is cleaner. I'll do the changes. Thanks.

mariia-aloshyna (approved):

PR #2355, opened 2023-11-24
UIIN-2699 Make browse result items that are not anchors and have no records not clickable, and show 0 in number of titles.

BogdanDenis:
## Description
Make browse result items that are not anchors and have no records not clickable, and show 0 in number of titles.

This is to fix a bug with Browsing contributors and subjects. In consortial environment, when browsing with `Shared: No` back-end returns items related to shared records. This is because search index does not differentiate between local and shared Instance records.
Since we can't remove the items from results without changing the index, the workaround for now is to identify such records (non-anchor and with 0 total results) and make them not clickable

## Screenshots

https://github.com/folio-org/ui-inventory/assets/19309423/c236a1f6-e0ff-4fd0-b9e0-338c39f1e4d5

## Issues
[UIIN-2699](https://issues.folio.org/browse/UIIN-2699)

### DIFF
--- a/src/components/BrowseResultsList/getBrowseResultsFormatter.test.js
+++ b/src/components/BrowseResultsList/getBrowseResultsFormatter.test.js
@@ -151,9 +151,16 @@ describe('getBrowseResultsFormatter', () => {
         contributorTypeId: ['contributorTypeId'],
         contributorNameTypeId: 'contributorNameTypeId',
         totalRecords: 1,
+      },
+      {
+        name: 'Antoniou, Grigoris 2',
+        contributorTypeId: ['contributorTypeId'],
+        contributorNameTypeId: 'contributorNameTypeId',
+        totalRecords: 0,
+        isAnchor: false,
       }
     ];
-    const [anchorRecord, nonAnchorRecord] = contentData;
+    const [anchorRecord, nonAnchorRecord, notClickableRecord] = contentData;
 
     const renderContributorsList = (params = {}) => renderComponent({
       contentData,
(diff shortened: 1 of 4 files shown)


src/components/BrowseResultsList/getBrowseResultsFormatter.test.js
```diff
@@ -171,6 +178,9 @@ describe('getBrowseResultsFormatter', () => {
       // Default row
       expect(screen.getByText(nonAnchorRecord.name).tagName.toLowerCase()).not.toBe('strong');
       expect(screen.getByText(nonAnchorRecord.totalRecords).tagName.toLowerCase()).not.toBe('strong');
+      // Non clickable row
+      expect(screen.getByText(notClickableRecord.name).tagName.toLowerCase()).not.toBe('strong');
```
mariia-aloshyna, line 182: `strong` tag doesn't mean that the element is not clickable, it's better to check that the element doesn't have `href` attribute for example.

BogdanDenis: This tests just checks that this item was rendered. There is a second test that checks that the item is not clickable and does not redirect to another page

mariia-aloshyna: Then why do you need to check that tag name is not strong? To check that the item was rendered will be enough to use `toBeInTheDocument()`. I don't quite understand this `strong` checking, is this header or something?

Dmytro-Melnyshyn (approved):

src/components/BrowseResultsList/getBrowseResultsFormatter.test.js
BogdanDenis: It shouldn't be `strong` because it's not a full match result and doesn't have any associated records

PR #2361, opened 2023-11-27
UIIN-2697: Correctly display location data for holdings and items when tenant is changed

OleksandrHladchenko1:
## Purpose

* When user on Member-1 trying to open details of Holdings/item from Member-2 Permanent location for Holding and effective location for Item displaying "Inactive" instead of the name of location 

## Approach

* Fetch fresh locations for tenant in `ViewHoldingRoute` component and pass locations to `ViewHoldingsRecord` component
* Fetch fresh locations for tenant in `ItemRoute` component and pass locations to `ItemView` component
* Updated unit tests

## Refs

[UIIN-2697](https://issues.folio.org/browse/UIIN-2697)

## Screenshots

https://github.com/folio-org/ui-inventory/assets/85172747/19b62385-a2d8-44a5-8c56-30a8ed3a950d

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * Don't render Fast Add record modal in a `<Paneset>` to re-calculate other `<Pane>`'s widths after closing. Fixes UIIN-2690.
 * "Saving instance failed" modal does not show error message. Fixes UIIN-2686.
 * Make browse result items that are not anchors and have no records not clickable, and show 0 in number of titles. Fixes UIIN-2699.
+* Browse Lists | Focus updates. Fixes UIIN-2267.
 * Correctly display location data for holdings and items when tenant is changed. Fixes UIIN-2697.
 
 ## [10.0.6](https://github.com/folio-org/ui-inventory/tree/v10.0.6) (2023-11-24)
--- a/src/components/BrowseResultsPane/BrowseResultsPane.js
+++ b/src/components/BrowseResultsPane/BrowseResultsPane.js
@@ -26,6 +26,7 @@ const BrowseResultsPane = ({
   isFetching,
   isFiltersOpened,
   pagination,
+  paneTitleRef,
   toggleFiltersPane,
   totalRecords,
 }) => {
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 * Ignored hot key command on edit fields. Refs UIIN-2604.
 * Don't render Fast Add record modal in a `<Paneset>` to re-calculate other `<Pane>`'s widths after closing. Fixes UIIN-2690.
 * "Saving instance failed" modal does not show error message. Fixes UIIN-2686.
+* Inactive permanent location for Holdings/items when user open details. Fixes UIIN-2697.
```
mariia-aloshyna, line 15: Looks like it should be `10.0.7`

src/views/ItemView.js
mariia-aloshyna, line 1: As far as `referenceTable` is taken from the DataContext provider and there are other properties that refer to this data, will it make sense to check all of them, not only locations? If some of them are the same for all tenants then it's ok to leave it constructed from `referenceTable`, but if they are set individually for each tenant (may be set in settings and so on), then I guess we need to fetch data for different tenants individually, or refactor provider.

CHANGELOG.md
zburke: I agree with the original implementation: the target branch here is master, so I think `11.0.0` is the right place right now. If we decide to cherry-pick this on to the `b10.0` branch and publish another patch release, _then_ `10.0.7` will be correct.

zburke (commented):
The whole point of stuffing `referenceTables` into context is to be able to write things like
```
const holdingsPermanentLocation = referenceTables?.locationsById[holdingsRecord?.permanentLocationId];
```
and have it just work. 1) I don't understand why that data is insufficient. 2) I don't understand why, noticing that it is insufficient, this was fixed by calling an additional hook and adding an additional prop instead of just fixing up the property that already exists. 

Additionally, the PR description and CHANGELOG entry should describe the solution instead of restating the problem. A good way to write a PR title is to complete the sentence, "After this PR merges, the code will ...", e.g. "Correctly display consortia location data for holdings and items"

ryandberger (approved):

OleksandrHladchenko1:
@zburke @mariia-aloshyna Please, review new changes

mariia-aloshyna (approved):

PR #2366, opened 2023-12-06
UIIN-2716 Refactor CSS away from postcss-color-function

JohnC-80:
https://issues.folio.org/browse/UIIN-2716

We're on a mission to remove an out-of-date dependency of our CSS pipeline, `postcss-color-function`. `color()` as we used it, was removed from its proposal and ultimately changed in native CSS API to convert color spaces rather than blend colors. 

The functionality has been replaced via the `color-mix()` function as well as the new relative-color-syntax. This PR uses both.

The colors are a very near match..
image

This relates back, but is only blocked by https://github.com/folio-org/stripes-webpack/pull/133 for firefox support.


mkuklis (approved):

BogdanDenis:
Thanks @JohnC-80, looks good!
There is one failing test due to timeout, I'll try to find a solution to fix it

PR #2371, opened 2023-12-10
UIIN-2559: The quantity of assigned tags to the instance is not displaying in the little tag icon after assigning

mariia-aloshyna:
## Purpose

The quantity of assigned tags to the instance is not displaying in the a little tag icon after assigning.

## Approach

- Refetch instance after successful mutation. The bug was caused by refactoring the instance details pane to use `react-queries`.

## Refs

https://issues.folio.org/browse/UIIN-2559


OleksandrHladchenko1 (commented):
Please, fix eslint warnings

BogdanDenis (approved):

PR #2373, opened 2023-12-15
UIIN-2715: Reset search/browse query when user switches affiliation.

Dmytro-Melnyshyn:
## Purpose
Reset search/browse query when user switches affiliation.
Remove the `useLogout` hook, due to having a callback for the `LOGIN` event.

## Related PRs
https://github.com/folio-org/ui-consortia-settings/pull/101
https://github.com/folio-org/stripes-core/pull/1383

## Refs
[UIIN-2715](https://issues.folio.org/browse/UIIN-2715)

## Screenshots

https://github.com/folio-org/ui-inventory/assets/77053927/c73eef62-ad5c-4942-96b4-528e47b5f750


src/providers/LastSearchTermsProvider.js
```diff
@@ -6,26 +6,11 @@ import { useNamespace } from '@folio/stripes/core';
 import { LastSearchTermsContext } from '../contexts';
 import { getItem, setItem } from '../storage';
 import { INIT_PAGE_CONFIG } from '../hooks/useInventoryBrowse';
-import { useLogout } from '../hooks';
 import { segments } from '../constants';
 
 const LastSearchTermsProvider = ({ children }) => {
   const [namespace] = useNamespace();
 
-  const resetStorageAfterLogout = () => {
```
BogdanDenis, line 15: Why is this function removed if we need to clear search term and results when logging out/switching affiliation?

src/index.js
```diff
@@ -198,6 +201,12 @@ const InventoryRouting = (props) => {
   );
 };
 
+InventoryRouting.eventHandler = (event) => {
+  if ([coreEvents.LOGIN, 'SWITCH_ACTIVE_AFFILIATION'].includes(event)) {
```
BogdanDenis, line 205: Let's make `SWITCH_ACTIVE_AFFILIATION` a constant

src/index.js
```diff
@@ -198,6 +201,12 @@ const InventoryRouting = (props) => {
...
+    sessionStorage.clear();
```
BogdanDenis, line 206: It's dangerous to clear all storage on login. Its purpose is to store data between page visits and if we clear it - it defeats the purpose and potentially introduces more bugs
With this event handler approach because there is just one handler per module - we need to somehow propagate these events to other parts of the app

BogdanDenis (changes requested):

src/providers/LastSearchTermsProvider.js
Dmytro-Melnyshyn: This function was added to avoid having search terms after a user logs out and then logs back in ([Scenario 1.1](https://issues.folio.org/browse/UIIN-2359)). By listening to the `LOGIN` event we get the same result.

src/index.js
Dmytro-Melnyshyn: > It's dangerous to clear all storage on login. Its purpose is to store data between page visits and if we clear it - it defeats the purpose and potentially introduces more bugs.

We need to avoid having the search terms after logout then login and after switching affiliation, so I'm going to clear only search terms in the next commit.

> With this event handler approach because there is just one handler per module - we need to somehow propagate these events to other parts of the app

We can process these events directly here, and if any case occurs, we can create an EventEmitter.

BogdanDenis (approved):

Terala-Priyanka (changes requested):
Could you please fix the linting errors?

PR #2376, opened 2023-12-24
UIIN-2728: Show locations from other tenants in Effective location facet.

Dmytro-Melnyshyn:
## Purpose
- show locations from other tenants in the Effective location facet

## Issues

[UIIN-2728](https://issues.folio.org/browse/UIIN-2728)

## Screenshots

https://github.com/folio-org/ui-inventory/assets/77053927/7fa6d1cd-ba20-45bb-8c9e-b5814b507e39

### DIFF
--- a/src/components/EffectiveLocationFacet/EffectiveLocationFacet.js
+++ b/src/components/EffectiveLocationFacet/EffectiveLocationFacet.js
@@ -1,0 +1,88 @@
+import { useContext } from 'react';
+import { useIntl } from 'react-intl';
+import PropTypes from 'prop-types';
+
+import { useStripes } from '@folio/stripes/core';
+import {
+  Accordion,
+  FilterAccordionHeader,
+} from '@folio/stripes/components';
+
+import CheckboxFacet from '../CheckboxFacet';
+import DataContext from '../../contexts/DataContext';
+import { useLocationsOfAllTenantsQuery } from '../../hooks';
+import { getFacetOptions } from '../../facetUtils';
+import { isUserInConsortiumMode } from '../../utils';
+
+const EffectiveLocationFacet = ({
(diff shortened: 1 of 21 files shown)


src/components/EffectiveLocationFacet/EffectiveLocationFacet.js
```diff
@@ -0,0 +1,88 @@
...
+  onFilterSearch,
+}) => {
+  const stripes = useStripes();
+  const intl = useIntl();
+  const { allTenantIds } = useContext(DataContext);
+
+  const {
+    data: locationsOfAllTenants,
+    isLoading: isLoadingAllLocations,
+  } = useLocationsOfAllTenantsQuery({ tenantIds: allTenantIds, enabled: open });
```
BogdanDenis, line 36: For facets in Inventory app we can't simply send a request to `/locations` and use the response as options because the response doesn't have facet counts.
And it doesn't fix the root cause - `/facets` endpoint returns all locations, but we're only requesting locations from current tenant in `DataProvider` so `getFacetOptions` can't find a match by id.
To fix this we need to use `useLocationsOfAllTenantsQuery` inside `DataProvider` and leave facets as-is.

Dmytro-Melnyshyn: > For facets in Inventory app we can't simply send a request to /locations and use the response as options because the response doesn't have facet counts.

/facets response returns ids, we use those ids to find names of options in /locations response.

> And it doesn't fix the root cause - /facets endpoint returns all locations, but we're only requesting locations from current tenant in DataProvider so getFacetOptions can't find a match by id.

I think the issue is that the locations of other tenants should be shown too.

PR #2377, opened 2023-12-26
UIIN-2692: Moving items between holdings records doesn't update correctly in UI

mariia-aloshyna:
## Purpose

When moving items between holdings records on the same instance, the UI does not update, so the items appear in both holdings records until the instance detail pane is closed and reopened.

## Approach

- Set empty array to items component state when `items` is empty.

## Refs

https://issues.folio.org/browse/UIIN-2692


src/Instance/ItemsList/ItemsListContainer.js
```diff
@@ -42,9 +42,9 @@ const ItemsListContainer = ({
   const { totalRecords } = useHoldingItemsQuery(holding.id, { searchParams: { limit: 0 }, key: 'itemCount', tenantId });
 
   useEffect(() => {
-    if (!isEmpty(items)) {
-      setItemsToShow(items);
-    }
+    const updatedItemsList = !isEmpty(items) ? items : [];
```
Terala-Priyanka, line 45: Looks like items is always an array either with length 0 or greater than 0 if there are items in holdings.
hence, do you need this isEmpty check on items? You should probably update the state when new items are available.

mariia-aloshyna: @Terala-Priyanka `isEmpty` also checks for `undefined` and `null`, so if `items` is undefined nothing crushes. I need to update the state even in the case when all items have been moved to another holding and it's empty.

mariia-aloshyna:

src/Instance/ItemsList/ItemsListContainer.js
Terala-Priyanka: @mariia-aloshyna 
Looks like API throws an array always. Do you see any case when a null is sent through API?
As long as it is array. we don't need to have this isEmpty check. directly update the state with new response.

mariia-aloshyna: @Terala-Priyanka updated

Dmitriy-Litvinenko (approved):

PR #2393, opened 2024-01-29
UIIN-2740 Add 'Display summary' field to item enumeration data accordion

usavkov-epam:
## Purpose

Ensure the synchronized data points between different UIs (basically receiving and inventory) are named consistently.

## Approach
Add a new field "Display summary" to the item view and form.
Update "Caption" to "Display summary" in the holding receiving history.

## Refs

UI Jira - https://issues.folio.org/browse/UIIN-2740
BE Jira - https://issues.folio.org/browse/MODORDERS-1000
## Screenshots

https://github.com/folio-org/ui-inventory/assets/88109087/eeb1fac0-7166-47cd-ad17-25201abb8cc9

### DIFF
--- a/src/edit/items/ItemForm.js
+++ b/src/edit/items/ItemForm.js
@@ -593,6 +593,18 @@ class ItemForm extends React.Component {
                     <Row>
                       <Col sm={3}>
                         <Field
+                          label={<FormattedMessage id="ui-inventory.displaySummary" />}
+                          name="displaySummary"
+                          id="additem_displaySummary"
+                          component={TextField}
+                          rows={1}
+                          fullWidth
+                        />
+                      </Col>
+                    </Row>
+                    <Row>
+                      <Col sm={3}>
+                        <Field
                           label={<FormattedMessage id="ui-inventory.enumeration" />}
                           name="enumeration"
                           id="additem_enumeration"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
 * Add Set record for deletion option in Actions menu with Confirmation modal. Refs UIIN-2594.
 * Use `onSave` prop for quickMARC to handle saving records separately. Refs UIIN-2743.
 * Set Inventory settings HTML page title this format - `<<App name>> settings - <<selected page name>> - FOLIO`. Refs UIIN-2713.
+* Add "Display summary" field to item enumeration data accordion. Refs UIIN-2740.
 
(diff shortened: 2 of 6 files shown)


ryandberger (approved):

SerhiiNosko (approved):

NikitaSedyx (approved):

src/edit/items/ItemForm.js
```diff
@@ -590,6 +590,18 @@ class ItemForm extends React.Component {
                     id="acc03"
                     label={<FormattedMessage id="ui-inventory.enumerationData" />}
                   >
+                    <Row>
+                      <Col sm={3}>
+                        <Field
+                          label={<FormattedMessage id="ui-inventory.displaySummary" />}
+                          name="displaySummary"
```
zburke, line 597: @usavkov-epam, I'm confused. Here in ItemForm and in ItemView you _added_ the `displaySummary` field, but in HoldingReceivingHistory you _replaced_ the `caption` field with `displaySummary`, so that wasn't merely "Update the label of the caption field" (the change on line 20) but "Change which field is displayed" (lines 28, 30, 39). 

Should there have been updates to the related Okapi interface version that includes this new field? It used to be the case that calling an API with a field it didn't support would cause a 4xx error, so if this is a new field then (regrettably) adding a new field -- even an optional one -- is technically a breaking change since including it in a request to the previous version will now cause an error when it used to succeed.

In other words: either way, if `caption` was replaced by `displaySummary` or if `displaySummary` was added, I think there should have been a version change to an Okapi interface here. Was it already added in another PR or did we miss that?

usavkov-epam: Thank you for your comment @zburke!

So, changing the column's label for the receiving history of holding is related to [changes](https://github.com/folio-org/mod-orders/pull/826/files) in the "pieces" interface, which was upgraded to a new major version. For this change, we have another PR that upgrades the interface version in the `package.json` to a new major one.

As for the `displaySummary` field in the item view/edit, it's a new one provided by the "inventory" interface, that has minor updates to include this new field in the schema.

zburke: Thank you for the additional details about PR #2395, @usavkov-epam. Can we confirm that a PUT request to a `pieces` `2.x` endpoint with `displaySummary` in the body will succeed?

usavkov-epam: Oh, it was changed to `2.0 3.0` in #2395...
Thank you for noticing that.

> Can we confirm that a PUT request to a pieces 2.x endpoint with displaySummary in the body will succeed?

Nah, it is supposed to throw an error.

cc @alisher-epam

PR #2402, opened 2024-02-20
UIIN-2663: Jest/RTL: Cover `CreateHolding` component with unit tests

OleksandrHladchenko1:
## Refs

[UIIN-2663](https://folio-org.atlassian.net/browse/UIIN-2663)

## Screenshots

image

### DIFF
--- a/src/Holding/CreateHolding/CreateHolding.test.js
+++ b/src/Holding/CreateHolding/CreateHolding.test.js
@@ -1,0 +1,80 @@
+import '../../../test/jest/__mock__';
+
+import { MemoryRouter } from 'react-router-dom';
+import {
+  QueryClient,
+  QueryClientProvider,
+} from 'react-query';
+import {
+  render,
+  screen,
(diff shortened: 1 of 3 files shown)


UladzislauKutarkin (approved):

src/Holding/CreateHolding/CreateHolding.test.js
```diff
@@ -0,0 +1,80 @@
+import '../../../test/jest/__mock__';
```
artem-blazhko, line 1: This line is redundant. Could you remove it?
All set up of mocks import is implemented here - https://github.com/folio-org/ui-inventory/blob/master/jest.config.js

Dmitriy-Litvinenko, line 1: Could you please clarify do we need this import? We already define mock in  https://github.com/folio-org/ui-inventory/blob/master/test/jest/jest.setup.js#L1

artem-blazhko (approved):

PR #2404, opened 2024-02-22
UIIN-2596  Staff suppress facet - use `No` as default value. Hide the facet when users don't have permissions to use it.

BogdanDenis:
## Description
Apply `staffSuppress.false` filter to all search and facet requests by default.
Users without `Inventory: Enable staff suppress facet` permissions don't have the ability to search for staff suppressed records and to change selected Staff Suppress facet value.
Users with the permission can see and change facet values.

## Screenshots
**User with Staff Suppress facet permissions**

https://github.com/folio-org/ui-inventory/assets/19309423/f14fa8f4-ed04-4a19-971e-cb0520aac151

**User without Staff Suppress facet permissions**

https://github.com/folio-org/ui-inventory/assets/19309423/8282d79c-ed2f-41af-8687-145e596a87b7

## Issues
[UIIN-2596](https://folio-org.atlassian.net/browse/UIIN-2596)

### DIFF
--- a/src/components/InstanceFilters/InstanceFilters.js
+++ b/src/components/InstanceFilters/InstanceFilters.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 import PropTypes from 'prop-types';
 import { FormattedMessage, useIntl } from 'react-intl';
 import _ from 'lodash';
@@ -9,6 +9,7 @@ import {
   FilterAccordionHeader,
 } from '@folio/stripes/components';
 import {
+  IfPermission,
   checkIfUserInMemberTenant,
   useStripes,
 } from '@folio/stripes/core';
@@ -185,6 +186,15 @@ const InstanceFilters = props => {
     props.data
   );
 
+  const isStaffSuppressFilterAvailable = stripes.hasPerm('ui-inventory.instance.view-staff-suppressed-records');
+
+  useEffect(() => {
+    onChange({
+      name: FACETS.STAFF_SUPPRESS,
+      values: ['false'],
+    });
+  }, []);
+
   const isUserInMemberTenant = checkIfUserInMemberTenant(stripes);
 
   return (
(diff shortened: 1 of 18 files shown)


src/components/InstanceFilters/InstanceFilters.js
```diff
@@ -185,6 +186,10 @@ const InstanceFilters = props => {
     props.data
   );
 
+  const isStaffSuppressFilterAvailable = stripes.hasPerm('ui-inventory.instance.view-staff-suppressed-records');
+
+  useStaffSuppressInitialValue(onChange, props.data.query);
```
Dmytro-Melnyshyn, line 191: Let's call the `onChange` callback with both the selected option and the staff suppress. This will allow us to make only one change in history and have only one request with the same endpoint to the server.

BogdanDenis: `onFilterChangeHandler` already merges staff suppress and other selected filters, so we don't have to do that anywhere else.
image

The first request is made by default when the app is mounted, and query string is built in `buildQuery` function
In that function we don't know if this is the very first request and we don't have staff suppress filter because it's not in the url, or we don't have it because a user has unchecked the option in the facet component.
We can only safely change filters in the function if search is empty or the user doesn't have the permission so we know for sure that adding staff suppress value if it's not there does not go agains user's actions

Dmytro-Melnyshyn: I'm referring having two instance requests and two facets requests after selecting the facet option. 

https://github.com/folio-org/ui-inventory/assets/77053927/14194f25-723f-48d8-b949-4e87d9f206c8

Dmytro-Melnyshyn: It is working now, thanks. 

I also think that we need to pass location into the hook instead of using it inside, since this hook will be used in the `find-instance` plugin and extracted to `stripes-inventory-componets`. There is no plugin's URL. 

It would be nice to change `history` via `replace` instead of `push` when we change filters automatically.

Dmytro-Melnyshyn (approved):

src/hooks/useStaffSuppressInitialValue/useStaffSuppressInitialValue.js
```diff
@@ -0,0 +1,38 @@
...
+  useEffect,
+  useRef,
+} from 'react';
+import { useLocation } from 'react-router-dom';
+
+import { FACETS } from '../../constants';
+
+const useStaffSuppressInitialValue = (onChange, query) => {
+  const location = useLocation();
+  const isQueryFiltersSet = useRef(Boolean(query.filters));
```
Dmytro-Melnyshyn, line 11: This can be simplified to `const isQueryFiltersSet = Boolean(query.filters);`. Therefore, some of the hooks can be removed.

PR #2405, opened 2024-02-26
UIIN-2778: Display Holdings/Item Electronic Access URIs consistent with Instances

OleksandrHladchenko1:
## Purpose

* There’s security vulnerability for holdings and items, but instances is fine.  The goal here is to adjust the code for holdings and items so it’s consistent with instances and therefore no longer a security risk.

## Approach

* Made Holdings/Item Electronic Access URIs consistent with Instances

## Refs

[UIIN-2778](https://folio-org.atlassian.net/browse/UIIN-2778)

### DIFF
--- a/src/ViewHoldingsRecord.js
+++ b/src/ViewHoldingsRecord.js
@@ -1075,15 +1075,22 @@ class ViewHoldingsRecord extends React.Component {
                           }}
                           formatter={{
                             'URL relationship': x => this.refLookup(referenceTables.electronicAccessRelationships, get(x, ['relationshipId'])).name || noValue,
-                            'URI': x => (get(x, ['uri'])
-                              ? (
-                                <a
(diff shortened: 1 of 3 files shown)


src/ViewHoldingsRecord.js
```diff
@@ -1075,15 +1075,22 @@ class ViewHoldingsRecord extends React.Component {
...
-                              ? (
-                                <a
-                                  href={get(x, ['uri'])}
-                                  style={wrappingCell}
-                                >
-                                  {get(x, ['uri'])}
-                                </a>)
-                              : noValue),
+                            'URI': x => {
+                              const uri = x?.uri;
```
UladzislauKutarkin, line 1079: can we simplify `return x?.uri`; ?

artem-blazhko (approved):

PR #2408, opened 2024-03-01
UIIN-2546/UIIN-2548/UIIN-2549/UIIN-2556/UIIN-2557  Number generator implementation

EthanFreestone:
## Purpose
Make use of the service interaction feature "number generator" to handle item barcode, accession number and call number fields in inventory.

This is an implementation to guide usage of the number generator library of components supplied by `ui-service-interaction`. This is not the _only_ way to use them, and instead represents a guide on how they _might_ be used, with the power of change living with the developer team, this is not a plugin over which you have 0 control, but rather an example implementation of this cross module functionality.

## Approach
This comprises of 3 main "feature blocks"

### Settings
New settings page - Number generator options.
This is a form containing two accordions, one for holding level interactions and one for item level interactions.

These set up simple checkboxes to turn on/off number generator for the fields in question, with 3 levels of configurability for the number generator:
- off (field exists as before)
- on without edit (field is disabled, the number generator button is rendered and is the only way to fill the field)
- on (field is non disabled, can either be filled in manually or by using the generator)

For items there is an additional ability - use accession number for call number. This will allow the user to specify that the accession number sequences are _also_ the options for call number field, and the two will share the same sequences and counts.

These settings currently make use of mod-configuration, but in reality so long as they can be stored and retrieved it does not matter how these settings are configured.

### Holdings form
Added number generator button to call number field in holdings form
### Item form
Added number generator button to call number, accession number and item barcode fields in items form


mariia-aloshyna:
Hi @EthanFreestone could you add the description to the PR, please? Like the purpose, approach, and maybe some screenshots?

EthanFreestone:
> Hi @EthanFreestone could you add the description to the PR, please? Like the purpose, approach, and maybe some screenshots?

I believe the purpose is covered by the linked JIRA tickets, a brief instruction on the approach is now part of the description. 

Effectively though - we've created a library of components which can make use of the "number generator" functionality in mod-service-interaction. This is an example of how inventory _might_ wish to implement that functionality, given the JIRA tickets raised.

The component library is highly configurable, and there are (as with all frontend work) many ways to skin this particular cat

src/settings/NumberGeneratorOptions/NumberGeneratorOptions.test.js
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+
+import { screen } from '@folio/jest-config-stripes/testing-library/react';
+import stripesFinalForm from '@folio/stripes/final-form';
```
mariia-aloshyna, line 4: There is `renderWithFinalForm` helper in ui-inventory, you can use it

src/settings/NumberGeneratorOptions/NumberGeneratorOptionsForm.js
mariia-aloshyna, line 1: Please, cover this component with unit tests

package.json
```diff
@@ -484,6 +485,14 @@
         ],
         "visible": true
       },
+      {
+        "permissionName": "ui-inventory.settings.numberGenerator.manage",
```
mariia-aloshyna, line 489: Add the new permission to translations as well

elsenhans:
See https://github.com/folio-org/ui-inventory/pull/2764

PR #2415, opened 2024-03-19
UIIN-2807: User with "Inventory: All permissions" permission only should not be seeing the "Staff suppress" facet

mariia-aloshyna:
## Purpose

User with "Inventory: All permissions" permission only should not be seeing the "Staff suppress" facet in “Inventory” app and in “Select Instance” plug-in.
Only user which has "Inventory: Enable staff suppress facet" UI permission should be seeing the "Staff suppress" facet.

## Approach

Exclude "Inventory: Enable staff suppress facet" perm from "Inventory: All permissions" subpermissions.

## Refs

https://folio-org.atlassian.net/browse/UIIN-2807


OleksandrHladchenko1 (approved):

mariia-aloshyna:
Hi @Dmytro-Melnyshyn @BogdanDenis please take a look

Dmytro-Melnyshyn (approved):

PR #2423, opened 2024-03-26
UIIN-2787: Do not remove field from the form when its value is an empty string

OleksandrHladchenko1:
## Purpose

* When create an electronic access entry in either **holdings** or **item** or **instance** record and erase field value or select the default value, `react-final-form` removes the field from the form. We need to keep the field in the form with value `''`

## Approach

* In order to allow form values of `''` we need to pass `v => v` function to `parse` prop

## Refs

[UIIN-2787](https://folio-org.atlassian.net/browse/UIIN-2787)
[Final Form - FieldProps - parse](https://final-form.org/docs/react-final-form/types/FieldProps#parse)

### Before
https://github.com/folio-org/ui-inventory/assets/85172747/39ca01ed-257d-4542-a76d-b1d5c2e92594

### After
https://github.com/folio-org/ui-inventory/assets/85172747/d7c43edd-bfd1-430f-856c-aec521434f74


BogdanDenis (approved):

src/utils.js
```diff
@@ -910,3 +910,5 @@ export const setRecordForDeletion = async (okapi, id, tenantId) => {
 
   return response;
 };
+
+export const parseEmptyFormValue = value => value;
```
mariia-aloshyna, line 914: Cover it with tests, please

PR #2425, opened 2024-03-27
UIIN-2816: Replace 'all' with '=' to fix the 'All' search option that was preventing contributors from being searched in MARC and FOLIO-source records.

Dmytro-Melnyshyn:
## Description
As a result of investigating the bug in [MSEARCH-606](https://folio-org.atlassian.net/browse/MSEARCH-606), it was decided that the `All` search option should use the `=` operator instead of `all` to support more efficient searches. The operator was modified for `Instance`, `Holdings` and `Item`, as well as for advanced searches.

## Refs
[UIIN-2816](https://folio-org.atlassian.net/browse/UIIN-2816)

## Screenshots

https://github.com/folio-org/ui-inventory/assets/77053927/f03c0813-fd3b-488f-896b-797bb537a5cb


BogdanDenis (approved):

zburke (commented):
There isn't enough information in either the ticket or the PR to evaluate this. Can you provide a link to the CQL documentation for `=` and `all`? 

The PR title indicates this change is necessary to get "correct" results, but then the description indicates this is to get "more efficient" results. Which is it, i.e. is `=` just faster, or is it different? If it's different, how? Can you provide before and after examples? i.e. sth like "all would match X, Y, Z, but = would only match X, Z because ..."

It isn't clear how the attached video is relevant since the underlying query is not displayed for any of the searches.

Dmytro-Melnyshyn:
> There isn't enough information in either the ticket or the PR to evaluate this. Can you provide a link to the CQL documentation for `=` and `all`?
> 
> The PR title indicates this change is necessary to get "correct" results, but then the description indicates this is to get "more efficient" results. Which is it, i.e. is `=` just faster, or is it different? If it's different, how? Can you provide before and after examples? i.e. sth like "all would match X, Y, Z, but = would only match X, Z because ..."
> 
> It isn't clear how the attached video is relevant since the underlying query is not displayed for any of the searches.

Hi @zburke - The `All` search option is not searching Contributor in either MARC or FOLIO-source records.
The BE team [investigated](https://folio-org.atlassian.net/browse/MSEARCH-606?focusedCommentId=205594) the issue and here is the result:
> Based on implementation in MSEARCH-134 the contributor name field is not multi-lang but standard type. It makes searches by contributor with deactivated fuzzy logic.
The ”All” search option in the query uses the “all” operator to a computed field that includes all multi-lang and not-multi-lang fields. Requests with the “all“ operator will not work with not-multi-lang fields when the input is a string with spaces. But it should work with the “=” operator.

This [video recording](https://github.com/folio-org/ui-inventory/assets/77053927/69c3aea9-3fc2-4e21-acef-5709745c2e42) shows that records with the contributor `Dobyns, Stephen, 1941-` could not be found. 
In [this video recording](https://github.com/folio-org/ui-inventory/assets/77053927/5ee4554c-6ffd-43d0-90db-2780dc6e38b4), records with the contributor `Dobyns, Stephen, 1941-` are found after changes made in the current PR.

Tomorrow I'll ask the BE team to update the mod-search [documentation](https://github.com/folio-org/mod-search?tab=readme-ov-file#cql-query-operators) related to the `=` and `all` operators.

mariia-aloshyna (approved):

zburke (changes requested):
Thanks for the updated documentation! Please make sure it gets into the git commit when you merge and not just left on the PR description.

Dmytro-Melnyshyn:
> Thanks for the updated documentation! Please make sure it gets into the git commit when you merge and not just left on the PR description.

@zburke I didn't update the documentation, the documentation is correct. In this PR, the search has been corrected in accordance with the documentation.

zburke (approved):
Code changes LGTM. 

Please don't force-push commits on an open PR. It destroys the context for the comments if the code is revised. You can keep a clean history by squashing when you merge.

PR #2435, opened 2024-04-08
UIIN-2814 Clear USER_TOUCHED_STAFF_SUPPRESS_STORAGE_KEY flag when switching between search segments (follow-up)

BogdanDenis:
## Description
We need to clear `USER_TOUCHED_STAFF_SUPPRESS_STORAGE_KEY ` when switching between search segments. If the flag is true we skip setting of default value for `staffSuppress` facet so we need to clear the flag.

## Screenshots

https://github.com/folio-org/ui-inventory/assets/19309423/1138cbc2-27c3-4905-9b2b-63759034583a

## Issues
[UIIN-2814](https://folio-org.atlassian.net/browse/UIIN-2814)

### DIFF
--- a/src/components/InstancesList/InstancesList.test.js
+++ b/src/components/InstancesList/InstancesList.test.js
@@ -207,6 +207,26 @@ describe('InstancesList', () => {
     });
 
     describe('when switching to Holdings tab', () => {
+      const mockSetItem = jest.fn();
+
+      beforeEach(() => {
+        global.Storage.prototype.setItem = mockSetItem;
+      });
+
+      afterEach(() => {
+        global.Storage.prototype.setItem.mockReset();
+      });
+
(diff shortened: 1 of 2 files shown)


Dmytro-Melnyshyn (approved):

src/components/InstancesList/InstancesList.test.js
```diff
@@ -207,6 +207,26 @@ describe('InstancesList', () => {
...
+
+      beforeEach(() => {
+        global.Storage.prototype.setItem = mockSetItem;
+      });
+
+      afterEach(() => {
+        global.Storage.prototype.setItem.mockReset();
+      });
+
+      it ('should clear USER_TOUCHED_STAFF_SUPPRESS_STORAGE_KEY', async () => {
```
Dmytro-Melnyshyn, line 220: Please fix eslint warnings.

vashjs (approved):

PR #2458, opened 2024-04-22
UIIN-2801: Split the 'containsAll' query by spaces for the 'Classification, Normalized' search option due to lack of BE support.

Dmytro-Melnyshyn:
## Description
Subsequent PR.

The purpose is to allow a user to use the `containsAll` option in advanced search for the `Classification, Normalized` search option.

If there are classification instances: 
- WA **675** 1967
- WA **I61w** 1967

The query `normalizedClassificationNumber any "*675 I61w*"` will not return both records because BE does not support the `any` operator for `normalizedClassificationNumber` due to its normalization. They skip all whitespace and special characters, so they can’t identify the parts of the input that should be searched by `any` operation. 

Splitting the query into smaller parts allows us to get correct results:
`normalizedClassificationNumber any "*675 I61w*"` =>`normalizedClassificationNumber any "*675*" or normalizedClassificationNumber any "*I61w*"`

## Refs
[UIIN-2801](https://folio-org.atlassian.net/browse/UIIN-2801)

## Screenshots

https://github.com/folio-org/ui-inventory/assets/77053927/6de44573-e7a4-4e13-81d0-effe2be91be8


BogdanDenis (approved):

artem-blazhko (commented):
Could you increase code coverage to required 80%?

mariia-aloshyna (approved):

PR #2461, opened 2024-04-24
UIIN-2666 Jest/RTL: Cover ConnectedTitle component with unit tests

vashjs:
After this PR merged: 

- coverage for ConnectedTitle will be 100%
- fixed a condition that was never met

Ref: [UIIN-2666](https://folio-org.atlassian.net/browse/UIIN-2666)

image

### DIFF
--- a/src/components/ConnectedTitle/ConnectedTitle.test.js
+++ b/src/components/ConnectedTitle/ConnectedTitle.test.js
@@ -1,0 +1,87 @@
+import React from 'react';
+import { BrowserRouter as Router } from 'react-router-dom';
+import { keyBy } from 'lodash';
+
+import { DataContext } from '../../contexts';
+import { identifierTypes, instanceRelationshipTypes } from '../../../test/fixtures';
+import ConnectedTitle from './ConnectedTitle';
+import renderWithIntl from '../../../test/jest/helpers/renderWithIntl';
+
+jest.mock('../InstancePlugin/InstancePlugin', () => ({ onSelect, onClose }) => {
+  return (
+    <div>
+      <div>InstancePlugin</div>
+      <button type="button" onClick={() => onSelect()}>onSelect</button>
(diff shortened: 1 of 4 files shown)


src/utils.js
```diff
@@ -474,6 +474,8 @@ export const getIdentifiers = (identifiers = [], type, identifierTypesById) => {
     }
   });
 
+  if (!result.length) return null;
```
vashjs, line 477: This change is required, because previous logic wasn't correct. In prev implementation we will never see Not available label, because isbn and issn always returns `true` from `getIdentifiers`.

UladzislauKutarkin (approved):

src/components/ConnectedTitle/ConnectedTitle.test.js
```diff
@@ -0,0 +1,87 @@
...
+    </DataContext.Provider>
+  </Router>
+);
+
+
+describe('ConnectedTitle', () => {
+  it('should render correct titles', () => {
+    const { queryByText } = renderConnectedTitle();
+
+    expect(queryByText('ui-inventory.precedingField.title')).toBeInTheDocument();
```
mariia-aloshyna, line 54: If you render the component with intl, shouldn't it render translations values instead of ids?

vashjs: Good point, thanks! I have added translations. To be honest, I like the option with keys more if we check what exactly was rendered and should not care about how it will be translated. I see that on the folio it's about 50/50, some just do keys, some add translations, but it looks like in inventory you do this with translations, so it makes sense

mariia-aloshyna: yes, I guess it makes sense to check values because we are actually testing UI and that's why we use react-testing-library, so it is necessary to check what a user sees on UI

PR #2469, opened 2024-04-28
UIIN-2664 Jest/RTL: Cover MoveHoldingContext component with unit tests

vashjs:
## Purpose

After this PR merged test coverage for MoveHoldingContext.js component will be ~80%
## Approach

Covered all functinality inside MoveHoldingContext except drag/drop between lists, because as per now it's not good way to do that for react-butiful-dnd. Covered moving of holindgs using dropdown menu instead.

## Refs

https://folio-org.atlassian.net/browse/UIIN-2664
## Screenshots

image

### DIFF
--- a/src/Instance/MoveHoldingContext/MoveHoldingContext.test.js
+++ b/src/Instance/MoveHoldingContext/MoveHoldingContext.test.js
@@ -1,0 +1,201 @@
+import React from 'react';
+import { BrowserRouter as Router } from 'react-router-dom';
+import { keyBy } from 'lodash';
+
+import { screen, waitFor, configure, fireEvent } from '@folio/jest-config-stripes/testing-library/react';
+import { within } from '@folio/jest-config-stripes/testing-library/dom';
+
+import { DataContext } from '../../contexts';
+import { useHoldings, useInstanceHoldingsQuery } from '../../providers';
+import { useLocationsQuery } from '../../hooks';
+import { holdingsById, identifierTypes, instanceRelationshipTypes } from '../../../test/fixtures';
+import renderWithIntl from '../../../test/jest/helpers/renderWithIntl';
+import translationsProperties from '../../../test/jest/helpers/translationsProperties';
+import MoveHoldingContext from './MoveHoldingContext';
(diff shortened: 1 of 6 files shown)


src/Instance/MoveHoldingContext/MoveHoldingContext.test.js
```diff
@@ -0,0 +1,201 @@
...
+    let grid;
+    let rows;
+
+    await waitFor(() => {
+      grid = within(annexSection).getByRole('grid');
+      rows = within(grid).getAllByRole('row');
+
+      expect(grid).toBeVisible();
+      expect(rows).toHaveLength(3);
+    });
```
mariia-aloshyna, line 146: Maybe it would be better to check by Holdings location names for example rather than by tags as far as we testing UI?

UladzislauKutarkin (approved):

src/Instance/MoveHoldingContext/MoveHoldingContext.test.js
```diff
@@ -0,0 +1,225 @@
...
+      clickMoveFlow({ getByText, getByTestId });
+
+      const confirmBtn = screen.getByRole('button', { name: /confirm/ });
+
+      fireEvent.click(confirmBtn);
+
+      expect(screen.queryByText('Loading')).toBeInTheDocument();
+    });
+
+    it('should close modal and stop moving when "Cancel" is clicked', async () => {
```
mariia-aloshyna, line 213: there are some extra `async` declarations left, also [here](https://github.com/folio-org/ui-inventory/pull/2469/files#diff-054a3ed1ef8fb650f861d0d464f2632a9911ee37779b23f3f4db12d096321ae9R122), and [here](https://github.com/folio-org/ui-inventory/pull/2469/files#diff-054a3ed1ef8fb650f861d0d464f2632a9911ee37779b23f3f4db12d096321ae9R201)

vashjs: I see, removed. My prev solution contained a lot of async data, but then I was able to avoid it, but I forgot remove all these async indicators

BogdanDenis (approved):

PR #2471, opened 2024-04-29
UIIN-2811 Use consolidated locations endpoint to fetch all locations when in central tenant context.

BogdanDenis:
## Description
 Use consolidated locations endpoint to fetch all locations when in central tenant context.

## Screenshots

https://github.com/folio-org/ui-inventory/assets/19309423/5118de48-895f-4413-919f-351a20b2a151

## Issues
[UIIN-2811](https://folio-org.atlassian.net/browse/UIIN-2811)

### DIFF
--- a/src/hooks/useLocationsForTenants/useLocationsForTenants.js
+++ b/src/hooks/useLocationsForTenants/useLocationsForTenants.js
@@ -1,9 +1,13 @@
-import { useQueries } from 'react-query';
+import {
+  useQueries,
+  useQuery,
+} from 'react-query';
 
 import {
   useNamespace,
   useOkapiKy,
   useStripes,
+  checkIfUserInCentralTenant,
 } from '@folio/stripes/core';
 
 import { isUserInConsortiumMode } from '../../utils';
@@ -18,9 +22,11 @@ const useLocationsForTenants = ({ tenantIds = [] }) => {
   const stripes = useStripes();
   const ky = useOkapiKy();
 
+  const isUserInCentralTenant = checkIfUserInCentralTenant(stripes);
+
   const queries = useQueries(tenantIds.map(tenantId => {
     return {
-      enabled: Boolean(tenantIds.length && isUserInConsortiumMode(stripes)),
+      enabled: Boolean(tenantIds.length && isUserInConsortiumMode(stripes) && !isUserInCentralTenant),
       queryKey: [namespace, 'locations', tenantId],
       queryFn: () => ky.get('locations', {
         searchParams: {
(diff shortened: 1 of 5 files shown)


Dmytro-Melnyshyn (approved):

src/hooks/useLocationsForTenants/useLocationsForTenants.js
```diff
@@ -38,10 +44,18 @@ const useLocationsForTenants = ({ tenantIds = [] }) => {
     };
   }));
 
-  const locationsFromAllTenants = queries.map(({ data }) => data?.locations).filter(Boolean).flat();
+  const { data: consolidatedLocations } = useQuery({
+    queryKey: [namespace, 'consolidatedLocations'],
+    queryFn: () => ky.get('search/consortium/locations').json(),
+    enabled: Boolean(isUserInConsortiumMode(stripes) && isUserInCentralTenant),
```
mariia-aloshyna, line 50: Why can't we use this endpoint on member tenants? This would improve performance (avoid sending 65 requests for 65 tenants for example) and also will allow us to display locations for Consortial holdings accordion for tenants with limited permissions

BogdanDenis: This is a limitation on back-end side unfortunately, the endpoint was implemented with this in mind

mariia-aloshyna: we use similar endpoints on member tenants - [useConsortiumHoldings](https://github.com/folio-org/ui-inventory/blob/master/src/hooks/useConsortiumHoldings/useConsortiumHoldings.js) and [useConsortiumItems](https://github.com/folio-org/ui-inventory/blob/master/src/hooks/useConsortiumItems/useConsortiumItems.js), but sending requests in the context of central tenant. Or are there some other limitations?

BogdanDenis: Maybe I misunderstood your question - `useConsortiumHoldings` and `useConsortiumItems` send requests when a user is in central tenant. Same with the locations endpoint. We only use it when user is in central tenant. If it's a member tenant then we should send multiple requests to a regular `/locations` endpoint
I think it's the same limitation as for holdings and items - user in a member tenant may not have permissions in central or other tenants to fetch locations from other tenants

mtraneis: Is there a situation where we have an admin from a member tenant who is affiliated with all other tenants in the consortia?

mariia-aloshyna: no-no, `useConsortiumHoldings` and `useConsortiumItems` are also sent from member tenants but in the context of the central tenant (x-okapi-tenant: central). 1 difference here is that they are sent only with the condition that a particular member tenant does not have permission to get holdings/items from inventory storage. You can find more [here](https://github.com/folio-org/ui-inventory/blob/4a86d95ec0bb7430d7e443d91b2ab21cda76286f/src/hooks/useMemberTenantHoldings/useMemberTenantHoldings.js#L8)

BogdanDenis: This endpoint still requires central tenant permissions, and in our case we want users without central tenant permissions to fetch locations from other tenants.

PR #2475, opened 2024-05-06
UIIN-2669: Jest/RTL: Cover ModalContent components with unit tests

UladzislauKutarkin:
https://folio-org.atlassian.net/browse/UIIN-2669
Here we added test coverage for `ModalContent.js `component
image


Terala-Priyanka (approved):

src/components/ModalContent/ModalContent.test.js
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
```
BogdanDenis, line 1: Please re-order imports:
1. Third-party imports
2. Imports from "@folio/..."
3. Imports from module

Also add an empty line separating each import group

src/components/ModalContent/ModalContent.test.js
```diff
@@ -0,0 +1,54 @@
...
+const itemMock = {
+  title: 'Sample Title',
+  barcode: '123456789',
+  materialType: {
+    name: 'Book',
+  },
+};
+
+describe('ModalContent', () => {
+  it('renders with correct props and messages', () => {
```
BogdanDenis, line 21: Let's not test 3 different cases in one `it`

Please split this test into multiple cases:
1. The modal content is rendered
2. when the cancel button is pressed the `onCancel` callback is called
3. when the confirm button is pressed the `onConfirm` callback is called

For this it will be useful to move
```
renderWithIntl(
      <MemoryRouter>
        <ModalContent
          stripes={stripesMock}
          item={itemMock}
          status={status}
          requestsUrl={requestsUrl}
          onCancel={onCancelMock}
          onConfirm={onConfirmMock}
          itemRequestCount={itemRequestCount}
        />
      </MemoryRouter>
    );
```
into a separate `renderModalContent` function

BogdanDenis (changes requested):

src/components/ModalContent/ModalContent.test.js
UladzislauKutarkin: Fixed!

src/components/ModalContent/ModalContent.test.js
UladzislauKutarkin: Thanks for that! Done!

BogdanDenis (approved):

PR #2479, opened 2024-05-09
UIIN-2896 Inventory app: Define and implement shortcut key for editing a quickMARC bib record

BogdanDenis:
## Description
Added new `Ctrl+Shift+E` shortcut to open Instance/Holdings records in quickMARC editor

## Screenshots

https://github.com/folio-org/ui-inventory/assets/19309423/6cdad120-d065-4cb5-90ad-1d7696e607be

## Issues
[UIIN-2896](https://folio-org.atlassian.net/browse/UIIN-2896)


src/components/ViewSource/ViewSource.js
```diff
@@ -51,6 +64,30 @@ const ViewSource = ({
...
   const [isMarcLoading, setIsMarcLoading] = useState(true);
 
+  const redirectToMARCEdit = useCallback(() => {
+    const urlId = isHoldingsRecord ? `${instanceId}/${holdingsRecordId}` : instanceId;
+    const pathname = `/inventory/quick-marc/edit-${isHoldingsRecord ? 'holdings' : 'bib'}/${urlId}`;
+
+    const searchParams = new URLSearchParams(location.search);
+
+    searchParams.delete('relatedRecordVersion');
+    searchParams.append('shared', instance.shared?.toString());
```
Dmytro-Melnyshyn, line 74: This is the second place we push `isShared` to URL. It might be worth to make a separate function for this. 
I think that if we open a locally shared record using the shortcut, we will push `isShared` value to false. If we are going to fix that behavior like 
```
const isInstanceShared = Boolean(isShared || isInstanceShadowCopy(instance?.source));
```
 then it would be nice to do it in just one place.

src/components/ViewSource/ViewSource.js
```diff
@@ -51,6 +64,30 @@ const ViewSource = ({
...
+
+    history.push({
+      pathname,
+      search: searchParams.toString(),
+    });
+  }, [isHoldingsRecord]);
+
+  const shortcuts = useMemo(() => [
```
Dmytro-Melnyshyn, line 82: Please add a second argument to `useMemo`.

Dmytro-Melnyshyn (approved):

UladzislauKutarkin (approved):

PR #2482, opened 2024-05-10
UIIN-2793: Populate Acquisitions accordion on instance when central ordering is active

OleksandrHladchenko1:
## Purpose

* Display Instance acquisitions for non-central (active) and central tenant.

## Approach

* If user is on **non-consortial** tenant or on **central** tenant - display only the list of instance acquisitions.
* If user is on **non-central** tenant - display **sub accordion** with the list of instance acquisitions for **active** and **central** tenant.
* Created `TenantAcquisition` component that is responsible only for rendering the list of instance acquisitions.
* Created and updated unit tests.

## Refs

[UIIN-2793](https://folio-org.atlassian.net/browse/UIIN-2793)

### Non-central tenant
image

### Central tenant
image

### Non-consortial tenant
image


src/Instance/InstanceDetails/InstanceAcquisition/TenantAcquisition.js
```diff
@@ -0,0 +1,58 @@
...
+  orderStatus: i => (
+    <>
+      {i.order?.workflowStatus ? <FormattedMessage id={`ui-inventory.acq.orderStatus.${i.order.workflowStatus}`} /> : <NoValue />}
+      {i.order?.orderCloseReason?.reason && ` - ${i.order.orderCloseReason.reason}`}
+    </>
+  ),
+  polReceiptStatus: i => <FormattedMessage id={`ui-inventory.acq.receiptStatus.${i.receiptStatus}`} />,
+  dateOrdered: i => getDateWithTime(i.order?.dateOrdered),
+  acqUnit: i => i.order?.acqUnits?.map(u => u.name)?.join(', ') || <NoValue />,
+  orderType: i => (i.order?.orderType ? <FormattedMessage id={`ui-inventory.acq.orderType.${i.order.orderType}`} /> : <NoValue />),
```
Terala-Priyanka, line 33: You should probably consider the column names picked from some constant.

src/Instance/InstanceDetails/InstanceAcquisition/InstanceAcquisition.js
```diff
@@ -1,66 +1,85 @@
...
+            label={getTenantAccordionLabel(stripes.user.user.tenants, centralTenant)}
+            className={css.tenantAcquisitionAccordion}
+            {...controlledCetralTenantAcqAccorion}
+          >
+            <TenantAcquisition
+              acquisitions={centralTenantAcquisition}
+              isLoading={isLoadingCentralTenantAcquisition}
+              tenantId={centralTenant}
+            />
+          </Accordion>
```
Terala-Priyanka, line 73: This one seems to be repetitive, could you try thinking of one of the below -
1. a pattern to render a component within an accordion? - may be a wrapper?
2. or conditionally rendering `TenantAcquisition` component within an accordion?

Terala-Priyanka (changes requested):
Please fix the failed unit tests.

src/constants.js
```diff
@@ -724,3 +724,12 @@ export const LEADER_RECORD_STATUSES = {
 };
 
 export const USER_TOUCHED_STAFF_SUPPRESS_STORAGE_KEY = 'folio_user_touched_staff_suppress';
+
+export const COLUMN_NAMES = {
```
Dmytro-Melnyshyn, line 728: - let's change the name to something more specific;
- this variable is used only in one place, maybe it shouldn't be added to general constants.

src/Instance/InstanceDetails/InstanceAcquisition/InstanceAcquisition.test.js
```diff
@@ -1,17 +1,30 @@
...
 import InstanceAcquisition from './InstanceAcquisition';
 import useInstanceAcquisition from './useInstanceAcquisition';
 
+import * as utils from '../../../utils';
+
+
+jest.mock('@folio/stripes/core', () => ({
+  ...jest.requireActual('@folio/stripes/core'),
+  useStripes: jest.fn(),
+}));
```
Dmytro-Melnyshyn, line 23: It is better to mock `useStripes` [here](https://github.com/folio-org/ui-inventory/blob/e0bc47eea5b2b1c365d5ed9e534a129e61c56615/test/jest/__mock__/stripesCore.mock.js#L86).

src/constants.js
OleksandrHladchenko1: I agree with the naming, but let's leave this variable in general constants, because it will be used in [`src/Holding/ViewHolding/HoldingAquisitions/HoldingAquisitions.js`](https://github.com/folio-org/ui-inventory/blob/e0bc47eea5b2b1c365d5ed9e534a129e61c56615/src/Holding/ViewHolding/HoldingAquisitions/HoldingAquisitions.js#L20) in scope of [UIIN-2817](https://folio-org.atlassian.net/browse/UIIN-2817)

Dmytro-Melnyshyn (approved):

BogdanDenis (approved):
Please add a CHANGELOG entry as well

src/Instance/InstanceDetails/InstanceAcquisition/InstanceAcquisition.js
```diff
@@ -1,66 +1,90 @@
...
-import {
-  Accordion,
-  MultiColumnList,
-  NoValue,
-} from '@folio/stripes/components';
+import { Accordion } from '@folio/stripes/components';
 
-import { getDateWithTime } from '../../../utils';
 import { useControlledAccordion } from '../../../common';
+import { isUserInConsortiumMode } from '../../../utils';
```
zburke, line 9: I know you didn't put `isUserInConsortiumMode` in `utils`, but it doesn't belong there; it belongs in `stripes-core` because there is _nothing_ inventory-specific about it. Note also that it is named incorrectly; `isUserInConsortiumMode` does not inspect the user, it inspects whether an interface is available on this tenant. 

Please think about this kind of thing when you encounter code like this: is it correct to do this work here, or should this be lifted up to a higher level? This will make stripes-core more useful and reduce code duplication by centralizing common features.

src/constants.js
```diff
@@ -724,3 +724,12 @@ export const LEADER_RECORD_STATUSES = {
 };
 
 export const USER_TOUCHED_STAFF_SUPPRESS_STORAGE_KEY = 'folio_user_touched_staff_suppress';
+
+export const ACQISITION_COLUMN_NAMES = {
```
zburke, line 728: Sorry, typo, please use `ACQUISITION`

zburke (approved):
Code changes LGTM. Please fix the typo in constants.js.

PR #2499, opened 2024-06-06
UIIN-2781: Translate search options.

Dmytro-Melnyshyn:
Follow-up PR.

## Purpose
Translate search options on the main page.

## Description
`formatSearchableIndex` was hardcoded to use `ui-inventory` to translate search options. The `prefix` variable is legacy code.

## Refs
[UIIN-2781](https://folio-org.atlassian.net/browse/UIIN-2781)

## Screenshots

image
image


BogdanDenis (approved):

Dmytro-Melnyshyn:
> This PR changes the signature of `formatSearchableIndex`. So ... don't we also need to change how we call it?

Yes, we need that, thanks for that. I renamed it to `translateSearchOptionLabel`.

zburke (approved):
Whoops, my bad, I didn't notice the change from `(index)` to `({ label, value })`, i.e. that the function signature isn't actually changing, rather, you're just destructuring the argument. Sorry!

Your pick on what the best name is :)

PR #2522, opened 2024-07-04
UIIN-2968 include translations from all shared @folio libraries

zburke:
Include `@folio/stripes-inventory-components` and
`@folio/stripes-marc-components` in `stripesDeps` to cause the translations defined in those shared libraries to be picked up by stripes-webpack when creating the bundle. Although it is possible to add translations to the bundle by directly including these modules in `stripes.config.js`, that's the wrong way to go about it (although it has the same effect) since these are just shared modules, not directly-useable applications.

Refs [UIIN-2968](https://folio-org.atlassian.net/browse/UIIN-2968)


BogdanDenis (approved):
Thanks for picking this up!

PR #2527, opened 2024-07-12
UIIN-2963: Add 'Edit in Linked Data Editor' option to Action drop down

s3fs:
https://folio-org.atlassian.net/browse/UIIN-2963


SKarolFolio (approved):

mariia-aloshyna (approved):

src/ViewInstance.js
```diff
@@ -932,6 +934,21 @@ class ViewInstance extends React.Component {
...
+          <MenuSection
+            id="ld-menu-section"
+            label={intl.formatMessage({ id: 'ui-inventory.ld.label' })}
+          >
+            <ActionItem
+              id="edit-resource-in-ld"
+              icon="edit"
+              messageId="ui-inventory.editInLinkedDataEditor"
+              onClickHandler={onToggle}
+              disabled={!canAccessLinkedDataOptions}
```
BogdanDenis, line 947: It's a pattern in Folio to hide buttons that can't be clicked due to missing permissions. Can we hide the menu action instead here?

BogdanDenis (changes requested):

src/ViewInstance.js
s3fs: The ticket states that the button needs to be present but disabled

BogdanDenis: Yes but requirements can be discussed and refined, and it's good to be consistend with how permissions are applied across other apps

s3fs: I've created a chat including you @BogdanDenis and our PO Doug Loynes so that you can propose this to him directly (I think he has some questions about that, too)

s3fs: Done @BogdanDenis

BogdanDenis (approved):

PR #2539, opened 2024-07-22
UIIN-2977 Save Instances UUIDs - When request URI exceeds character limit split request.

BogdanDenis:
## Description
Browsers have a limit of how many characters can be in a URI. When that limit exceeds HTTP requests fail
That happens in Inventory as well. When there are too many characters we need to split one query into multiple and join responses

## Screenshots

https://github.com/user-attachments/assets/b6253605-6454-4c92-9e47-c7148c0ef4a5

## Issues
[UIIN-2977](https://folio-org.atlassian.net/browse/UIIN-2977)

### DIFF
--- a/src/components/InstancesList/InstancesList.js
+++ b/src/components/InstancesList/InstancesList.js
@@ -8,6 +8,7 @@ import {
   size,
   isEmpty,
   noop,
+  uniqBy,
 } from 'lodash';
 import {
   injectIntl,
@@ -77,6 +78,7 @@ import {
   switchAffiliation,
   addFilter,
   replaceFilter,
+  batchQueryIntoSmaller,
 } from '../../utils';
 import {
   INSTANCES_ID_REPORT_TIMEOUT,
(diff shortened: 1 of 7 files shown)


src/components/InstancesList/InstancesList.js
```diff
@@ -872,7 +925,7 @@ class InstancesList extends React.Component {
             icon: 'save',
             messageId: 'ui-inventory.saveInstancesUIIDS',
             onClickHandler: buildOnClickHandler(this.generateInstancesIdReport),
-            isDisabled: isInstancesListEmpty,
+            isDisabled: false // isInstancesListEmpty,
```
Dmytro-Melnyshyn, line 928: It looks like it was a temporary solution.

BogdanDenis: Yes, I will use `isRequestUrlExceededLimit` from https://github.com/folio-org/ui-inventory/pull/2540/files#diff-0399c7f70f7645c1cfd8ef8c3b7f167b29a8dc385e86b129f0f46ac896a02144

Dmytro-Melnyshyn (approved):

artem-blazhko (approved):

PR #2541, opened 2024-07-24
UIIN-2978: Edit in LDE when source=MARC

s3fs:
https://folio-org.atlassian.net/browse/UIIN-2978

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,7 @@ and disable fields when "Settings (Inventory): Create, edit and delete HRID hand
 * Save Instances UUIDs - When request URI exceeds character limit split request. Refs UIIN-2977.
 * Use `withSearchErrors` HOC and `buildRecordsManifest` to display an error when the request URL is exceeded. Refs UIIN-2970.
 * Edit in LDE when source=MARC. Refs UIIN-2978.
+* Edit in Linked Data Editor when source=MARC. Refs UIIN-2978.
 
 ## [11.0.4](https://github.com/folio-org/ui-inventory/tree/v11.0.4) (2024-04-30)
 [Full Changelog](https://github.com/folio-org/ui-inventory/compare/v11.0.3...v11.0.4)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,7 @@ and disable fields when "Settings (Inventory): Create, edit and delete HRID hand
 * Display "New order" action in instance action menu when active affiliation set to central tenant. Refs UIIN-2955.
 * Save Instances UUIDs - When request URI exceeds character limit split request. Refs UIIN-2977.
 * Use `withSearchErrors` HOC and `buildRecordsManifest` to display an error when the request URL is exceeded. Refs UIIN-2970.
+* Edit in LDE when source=MARC. Refs UIIN-2978.
 
 ## [11.0.4](https://github.com/folio-org/ui-inventory/tree/v11.0.4) (2024-04-30)
 [Full Changelog](https://github.com/folio-org/ui-inventory/compare/v11.0.3...v11.0.4)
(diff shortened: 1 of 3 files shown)


src/ViewInstance.js
```diff
@@ -266,6 +275,7 @@ class ViewInstance extends React.Component {
 
     if ((isMARCSourceRecord || isLinkedDataSourceRecord) && (isViewingAnotherRecord || recordSourceWasChanged)) {
       this.getMARCRecord();
+      this.getLinkedDataEditorId();
```
pkjacob, line 278: When source = LINKED_DATA, there is no need to call the API for getting the linkedDataId. In such case, the linked data ID will be present in the `instance.identifiers` section of the instance record. 

Api has to be called only when source=MARC.

src/ViewInstance.js
```diff
@@ -236,6 +244,7 @@ class ViewInstance extends React.Component {
 
     if (isMARCSourceRecord || isLinkedDataSourceRecord) {
       this.getMARCRecord();
+      this.getLinkedDataEditorId();
```
pkjacob, line 247: When source = LINKED_DATA, there is no need to call the API for getting the linkedDataId. In such case, the linked data ID will be present in the `instance.identifiers` section of the `instance` record.

Api has to be called only when source=MARC.

src/ViewInstance.js
s3fs: Added a guard for when the source is not MARC

src/ViewInstance.js
s3fs: Added a guard for when the source is not MARC

BogdanDenis (approved):

src/ViewInstance.js
```diff
@@ -308,6 +318,25 @@ class ViewInstance extends React.Component {
...
+    const {
+      mutator,
+      stripes,
+      selectedInstance: { id, source },
+    } = this.props;
+
+    if (!id || !isMARCSource(source) || !stripes.hasPerm(LINKED_DATA_EDITOR_PERM)) return;
+
+    try {
+      const response = await mutator.linkedDataEditorId.GET(id);
```
pkjacob, line 331: I have one question.

When user navigate from one instance to the next instance, I guess the same ViewInstance component will be used for rendering the new Instance.

In such case, the "linkedDataEditorId" in the state will contain the linkedDataEditorId of the previously selected instance before the new API call returns(right?). If the user click on the link during that time, the user will be navigated to the edit screen of previously selected instance.

In order to avoid this, should we set the "linkedDataEditorId" in state as null before this line?

Or, am I totally wrong?

CHANGELOG.md
```diff
@@ -59,6 +59,7 @@ and disable fields when "Settings (Inventory): Create, edit and delete HRID hand
 * Add barcode values for "Bound pieces data" and update `useBoundPieces` hook query with `bindItemId` to fetch bound pieces correctly. Refs UIIN-2984.
 * Display "New order" action in instance action menu when active affiliation set to central tenant. Refs UIIN-2955.
 * Save Instances UUIDs - When request URI exceeds character limit split request. Refs UIIN-2977.
+* Edit in LDE when source=MARC. Refs UIIN-2978.
```
pkjacob, line 62: Let us mention "Linked Data Editor" instead of LDE

src/ViewInstance.js
pkjacob, line 331: I have one question.

When user navigate from one instance to the next instance, I guess the same ViewInstance component will be reused for rendering the new Instance.

In such case, the "linkedDataEditorId" in the state will contain the linkedDataEditorId of the previously selected instance before the new API call returns(right?). If the user click on the link during that time, the user will be navigated to the edit screen of previously selected instance.

In order to avoid this, should we set the "linkedDataEditorId" in state as null before this line?

Or, am I totally wrong?

mariia-aloshyna (approved):

pkjacob (approved):

PR #2545, opened 2024-07-30
UIIN-2995: Disable Move to button if no items selected or no more holdings exists within an instance

mariia-aloshyna:
## Purpose

Currently, if no items are selected during items movement within an instance, a user can still click on the Move to button, but the dropdown will be empty which is confusing, I suggest disabling the button in such case.

## Approach

- Disable the "Move to" button when movement is within the instance and there are no items selected or there are no holdings to move to
- Update tests

## Refs

https://folio-org.atlassian.net/browse/UIIN-2995

## Screenshots

active-move-to-button


BogdanDenis (approved):
Please fix the lint errors, otherwise looks good!

Dmytro-Melnyshyn (approved):

mariia-aloshyna:
> Please fix the lint errors, otherwise looks good!

I disagree with eslint, the escape character is necessary there, otherwise, the quote character will mean the end of the test description.

PR #2548, opened 2024-08-05
UIIN-2923 pass tenantId in InstancesRoute

NikitaSedyx:
## Purpose

in https://github.com/folio-org/stripes-inventory-components/pull/66 `tenantId` required prop was introduced

## Approach

add `tenantId` prop for `InstanceRoute` component


Dmytro-Melnyshyn (approved):

BogdanDenis (approved):
Thanks for fixing!

PR #2551, opened 2024-08-08
UIIN-2939: Edit Inventory Instances: "Save & close" and "Save and keep editing" buttons could be clicked 2 times in a row

mariia-aloshyna:
## Purpose

On the Edit Inventory Instances page:  "Save & close" and "Save and keep editing" buttons are not disabled right after first click.

## Approach

- Refetch instance so the initialValues are updated and form pristine is true

## Refs

https://folio-org.atlassian.net/browse/UIIN-2939


src/common/hooks/useInstance.js
```diff
@@ -37,6 +41,9 @@ const useInstance = (id) => {
     () => isSearchInstanceByIdLoading || isInstanceLoading,
     [isSearchInstanceByIdLoading, isInstanceLoading],
   );
+  const refetch = useCallback(() => {
+    refetchSearch().then(() => refetchInstance());
+  });
```
Dmytro-Melnyshyn, line 46: Can this be done in Promise.all?

Dmytro-Melnyshyn (approved):

src/common/hooks/useInstance.js
mariia-aloshyna: Promise.all execute promises in parallel, but we need to keep sequence of calls

PR #2566, opened 2024-08-19
UIIN-2889 Trim call numbers when Creating/Editing/Deriving Holdings/Items

BogdanDenis:
## Description
Trim call numbers when Creating/Editing/Deriving Holdings/Items.
New/updated Holdings/Items will not have trailing spaces

## Screenshots

https://github.com/user-attachments/assets/45ade92a-aa85-4a01-81d4-42cc8c7d6d79

## Issues
[UIIN-2889](https://folio-org.atlassian.net/browse/UIIN-2889)

### DIFF
--- a/src/edit/items/ItemForm.js
+++ b/src/edit/items/ItemForm.js
@@ -596,6 +596,8 @@ class ItemForm extends React.Component {
                           name="itemLevelCallNumberPrefix"
                           id="additem_callnumberprefix"
                           component={TextArea}
+                          format={v => v?.trim()}
+                          formatOnBlur
                           rows={1}
                           fullWidth
                         />
@@ -606,6 +608,8 @@ class ItemForm extends React.Component {
                           name="itemLevelCallNumber"
                           id="additem_callnumber"
                           component={TextArea}
+                          format={v => v?.trim()}
+                          formatOnBlur
                           rows={1}
                           fullWidth
                         />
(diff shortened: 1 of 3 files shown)


s3fs (approved):

src/edit/items/ItemForm.js
```diff
@@ -596,6 +596,8 @@ class ItemForm extends React.Component {
                           name="itemLevelCallNumberPrefix"
                           id="additem_callnumberprefix"
                           component={TextArea}
+                          format={v => v?.trim()}
+                          formatOnBlur
```
Dmytro-Melnyshyn, line 600: What if a user saves via a shortcut?

BogdanDenis: It also works with shortcuts

https://github.com/user-attachments/assets/51cccfff-c41c-4f30-8421-5ce2eb5554cb

BogdanDenis:
Hey @SKarolFolio thanks for review! I think because this function just calls a standard string method turning it into a util doesn't give much benefin

Dmytro-Melnyshyn (approved):

zburke (commented):
What is your goal here? You've added a [formatter](https://final-form.org/docs/react-final-form/types/FieldProps#format) but not a [parser](https://final-form.org/docs/react-final-form/types/FieldProps#parse), so this will trim call numbers on display but the actual submitted values will still contain leading/trailing whitespace.

PR #2567, opened 2024-08-19
UIIN-2939: Disable `Save and keep editing` and `Save & close` buttons when submit the instance form

OleksandrHladchenko1:
## Purpose

 * On the Edit Inventory Instances page:  "Save & close" and "Save and keep editing" buttons are not disabled right after first click. In order to avoid errors during saving the instance, we need to disable them.

## Approach

* Added `isSubmitting` prop to `InstanceForm` component that controls the state of both buttons.

## Refs

[UIIN-2939](https://folio-org.atlassian.net/browse/UIIN-2939)

## Screenshots

https://github.com/user-attachments/assets/9005a8b6-0bd3-497f-819d-51a2faefef99


zburke (approved):
LGTM, but I have a couple questions:
* this PR _adds_ `isSubmitting`, but it also removes the `submitting` prop. Why, and what are the consequences, if any? There is no mention of removing it in the PR description. 
* Can/should this be handled in a generic way for _all_ forms via stripes-form or stripes-final-form? It's fine if not, but it's at least worth considering whether this very generic feature could be handled in a more generic way.

mariia-aloshyna (approved):

OleksandrHladchenko1:
> LGTM, but I have a couple questions:
> 
> * this PR _adds_ `isSubmitting`, but it also removes the `submitting` prop. Why, and what are the consequences, if any? There is no mention of removing it in the PR description.
> * Can/should this be handled in a generic way for _all_ forms via stripes-form or stripes-final-form? It's fine if not, but it's at least worth considering whether this very generic feature could be handled in a more generic way.

Good catch! Thank you! After additional investigation I found out, that `submitting` prop was always `false` because `mutateInstance` was synchronous. I made `mutateInstance` asynchronous and now everything works as expected.

PR #2574, opened 2024-08-28
UIIN-2982: Rename Classification and Call numbers sections in browse options.

Dmytro-Melnyshyn:
## Purpose
Rename Classification and Call numbers sections in browse options.

## Refs
[UIIN-2982](https://folio-org.atlassian.net/browse/UIIN-2982)


BogdanDenis (approved):
Please check if these translations are used in e2e tests

mariia-aloshyna (approved):

PR #2579, opened 2024-09-02
UIIN-2982: Trigger translation update.

Dmytro-Melnyshyn:
## Refs
[UIIN-2982](https://folio-org.atlassian.net/browse/UIIN-2982)


BogdanDenis (commented):
Does changing of the order of keys trigger an update?
We can also just ask Peter Murray to trigger it without needing an additional PR, but if this works I don't have objections

Dmytro-Melnyshyn:
> Does changing of the order of keys trigger an update? We can also just ask Peter Murray to trigger it without needing an additional PR, but if this works I don't have objections

I think it works.

PR #2583, opened 2024-09-05
UIIN-2849 Instance Detail/Edit/New: Add new section for Date type, Date 1, & Date 2

BogdanDenis:
## Description
Instance records have new fields: Date Type, Date1 and Date2
Date Type is not required and can be saved empty
Date1 and Date2 are not required, but if a value is present it can only be exactly 4 characters long

## Screenshots

https://github.com/user-attachments/assets/96dbcf23-bd63-44a8-ab52-c2bccc457b77

## Issues
[UIIN-2849](https://folio-org.atlassian.net/browse/UIIN-2849)


src/providers/DataProvider.js
```diff
@@ -18,9 +21,10 @@ const DataProvider = ({
 
   const { commonData, isCommonDataLoading } = useCommonData();
   const { classificationBrowseConfig, isLoading: isBrowseConfigLoading } = useClassificationBrowseConfig();
+  const { instanceDateTypes, isLoading: isInstanceDateTypesLoading } = useInstanceDateTypes();
```
Dmytro-Melnyshyn, line 24: The date types will be used for the find-instance plugin too, let's move this hook to `useCommonData`.

mariia-aloshyna (approved):

Dmytro-Melnyshyn (approved):

PR #2589, opened 2024-09-11
UIIN-2932: User affiliation and service point information missing when affiliation changes in Inventory

mariia-aloshyna:
## Purpose

When a user selects a holding or item in Inventory, the affiliation changes, but the information (affiliation and service point) is no longer visible.

## Approach

- Update user (add tenants to stripes to display user affiliation) when switching affiliation.

## Refs

https://folio-org.atlassian.net/browse/UIIN-2932


BogdanDenis (approved):

src/utils.js
```diff
@@ -765,6 +785,7 @@ export const switchAffiliation = async (stripes, tenantId, move) => {
         perms: stripes.user.perms,
       },
     );
+    await handleUpdateUser(stripes);
```
Dmytro-Melnyshyn, line 788: This function contains three `await`s. Would combining them into `Promise.all` give the same result?

Dmytro-Melnyshyn (approved):

src/utils.js
Dmytro-Melnyshyn: The `switchAffiliation` function will complete its execution faster. `Promise.all` allows multiple asynchronous operations to be executed concurrently. Async/await waits for the previous one to complete before continuing.

Dmytro-Melnyshyn: Like this:
```
export const switchAffiliation = async (stripes, tenantId, move) => {
  if (stripes.okapi.tenant !== tenantId) {
    const updateTenantPromise = updateTenant(stripes.okapi, tenantId);
    
    const validateUserPromise = validateUser(
      stripes.okapi.url,
      stripes.store,
      tenantId,
      {
        token: stripes.okapi.token,
        user: stripes.user.user,
        perms: stripes.user.perms,
      },
    );
    
    const handleUpdateUserPromise = handleUpdateUser(stripes);
    
    await Promise.all([updateTenantPromise, validateUserPromise, handleUpdateUserPromise]);
    
    move();
  } else {
    move();
  }
 ```

mariia-aloshyna: thanks, changed to Promise.all

PR #2596, opened 2024-09-16
UIIN-3034 Implement default Date type when not present - No attempt to code

BogdanDenis:
## Description
When a user creates an Instance they can save dates without date type, but it should be saved as "No attempt to code"
If no dates are saved then date type should also be empty
When editing a record - date type cannot be unselected

## Screenshots

https://github.com/user-attachments/assets/1c7fe1c5-58f8-4bae-9db7-1b93e461c22e

## Issues
[UIIN-3034](https://folio-org.atlassian.net/browse/UIIN-3034)

### DIFF
--- a/src/edit/dateFields.test.js
+++ b/src/edit/dateFields.test.js
@@ -1,0 +1,66 @@
+import PropTypes from 'prop-types';
+
+import {
+  screen,
+  fireEvent,
+} from '@folio/jest-config-stripes/testing-library/react';
+import stripesFinalForm from '@folio/stripes/final-form';
+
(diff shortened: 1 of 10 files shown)


src/edit/dateFields.test.js
```diff
@@ -0,0 +1,66 @@
+import PropTypes from 'prop-types';
+
+import {
+  screen,
+  fireEvent,
```
artem-blazhko, line 5: Could you remove unused variable?

artem-blazhko (approved):

Dmytro-Melnyshyn (approved):

PR #2613, opened 2024-10-03
UIIN-3024: Disable link to POL when it leads to another tenant (instance, holding and item)

OleksandrHladchenko1:
## Purpose

* Link to PO line should be displayed in “Acquisition” accordion on instance, holding and item details pane. When PO line was created in tenant different from active, link to POL should be disabled.

## Approach

* For central tenant acquisition display po line number as a plain text.

## Refs

[UIIN-3024](https://folio-org.atlassian.net/browse/UIIN-3024)

## Screenshots

https://github.com/user-attachments/assets/bb59b252-bb42-403f-9c7d-8ebf3812b180


src/Item/ViewItem/ItemAcquisition/ItemAcquisition.js
```diff
@@ -17,14 +17,13 @@ import useItemAcquisition from './useItemAcquisition';
 
 const ItemAcquisition = ({ accordionId, itemId }) => {
   const stripes = useStripes();
-  const { isLoading, itemAcquisition = {} } = useItemAcquisition(itemId);
+  const { isLoading, itemAcquisition = {}, isCentralTenantAcq } = useItemAcquisition(itemId);
```
zburke, line 20: Pedantic: consider parity in how this boolean is named. Everywhere else it's `Acquisition` so why abbreviate it `Acq` here? Consistency makes it more obvious that these things are related and reduces the cognitive load.

zburke (approved):
Logic looks good but please fix the typo; use `Acqisition` should be `Acquisition`.

UladzislauKutarkin (approved):

PR #2614, opened 2024-10-04
UIIN-3054 Fetch bound item related pieces from the central tenant in the central ordering mode

usavkov-epam:
## Purpose

When a user is on a member tenant and central ordering is enabled in the platform, it's necessary to fetch bound item-related pieces from the active tenant and the central one.

## Refs

https://folio-org.atlassian.net/browse/UIIN-3054
## Screenshots

https://github.com/user-attachments/assets/5266b308-0418-4a50-b2c5-5f34b820615d

---

non-ECS

https://github.com/user-attachments/assets/6e29d635-c20c-4268-b7af-249abd47f27d

### DIFF
--- a/src/Instance/utils.js
+++ b/src/Instance/utils.js
@@ -1,3 +1,5 @@
+import { OKAPI_TENANT_HEADER } from '@folio/stripes-inventory-components';
+
 export const isItemsSelected = (items, selectedItemsMap) => {
   return items.every((item) => Boolean(
     selectedItemsMap[item.holdingsRecordId] && selectedItemsMap[item.holdingsRecordId][item.id]
@@ -54,6 +56,18 @@ export const navigateToHoldingsViewPage = (history, location, instance, holding,
     state: {
       tenantTo,
       tenantFrom,
+    },
+  });
+};
+
+export const extendKyWithTenant = (ky, tenantId) => {
+  return ky.extend({
+    hooks: {
+      beforeRequest: [
+        request => {
+          request.headers.set(OKAPI_TENANT_HEADER, tenantId);
+        },
+      ],
     },
   });
 };
--- a/src/common/hooks/useBoundPieces/useBoundPieces.js
+++ b/src/common/hooks/useBoundPieces/useBoundPieces.js
@@ -1,25 +1,42 @@
 import { useQuery } from 'react-query';
 
 import {
+  getConsortiumCentralTenantId,
   LIMIT_MAX,
   ORDER_PIECES_API,
+  useCentralOrderingSettings,
 } from '@folio/stripes-acq-components';
 import {
   useNamespace,
   useOkapiKy,
+  useStripes,
 } from '@folio/stripes/core';
 
+import { extendKyWithTenant } from '../../../Instance/utils';
+
 const DEFAULT_DATA = [];
+const SORTING_FIELD = 'receivedDate';
 
 const useBoundPieces = (itemId, options = {}) => {
   const { enabled = true, ...otherOptions } = options;
+
+  const stripes = useStripes();
   const ky = useOkapiKy();
   const [namespace] = useNamespace({ key: 'bound-pieces' });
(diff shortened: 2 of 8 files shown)


src/common/hooks/useBoundPieces/useBoundPieces.test.js
```diff
@@ -0,0 +1,95 @@
...
+
+describe('useBoundPieces', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+
+    useOkapiKy.mockReturnValue(mockKy);
+    useStripes.mockReturnValue(mockStripes);
+  });
+
+  it('should return default data when query is not enabled', async () => {
```
alisher-epam, line 63: `async` redundant

src/Instance/utils.js
```diff
@@ -57,3 +59,15 @@ export const navigateToHoldingsViewPage = (history, location, instance, holding,
...
+  return ky.extend({
+    hooks: {
+      beforeRequest: [
+        request => {
+          request.headers.set(OKAPI_TENANT_HEADER, tenantId);
+        },
+      ],
+    },
+  });
+};
```
zburke, line 73: What's the purpose of this? [`@folio/stripes/core::useOkapiKy()` already supports a `tenant` prop](https://github.com/folio-org/stripes-core/blob/0cdfe80aab2418acafb008650a3671babe1f0d15/src/useOkapiKy.js). Is this something different or am I misunderstanding? It's hard to tell what your goal is when there aren't any explanatory comments.

src/common/hooks/useBoundPieces/useBoundPieces.js
```diff
@@ -29,8 +46,34 @@ const useBoundPieces = (itemId, options = {}) => {
...
+      }
+
+      return makeRequest(ky);
+    },
+    enabled: Boolean(
+      enabled
+      && itemId
+      && consortiumCentralTenantId
+      && !isCentralOrderingSettingsFetching
+    ),
```
zburke, line 76: [stripes-acq-components::getConsortiumCentralTenantId()](https://github.com/folio-org/stripes-acq-components/blob/161e6644c91066a00ca80692894c49decc198079/lib/utils/consortia/getConsortiumCentralTenantId.js#L2) will return `undefined` for non-ECS tenants. Thus, after this change, this hook will only be enabled for ECS tenants. I don't think this is correct.[](url)

zburke (commented):
* I don't understand why you need ` extendKyWithTenant()` given `useOkapiKy()` will already let you override the tenant. This change looks unnecessary.
* I'm not confident `useBoundPieces()` still works correctly in non-ECS mode. This change looks incorrect.

src/common/hooks/useBoundPieces/useBoundPieces.js
usavkov-epam: Good catch, thanks! Will fix it

src/Instance/utils.js
usavkov-epam: Although `useOkapiKy` supports a passed tenant, it is still a hook and imposes restrictions for hooks. We need the ability to specify the tenant for the HTTP client dynamically from the received data, in this case a react hook will not help, except as will create a basic HTTP client, which we can configure.

E.g. we have to update a piece in a tenant where it's located.
https://github.com/folio-org/ui-inventory/pull/2614/files#diff-4a37a8dce3623a54155ee5e8a9e5845b7c2f1957ef1bad4c7e919aecdd15c36fR13

src/common/hooks/useBoundPieces/useBoundPieces.test.js
usavkov-epam: updated

src/common/hooks/useBoundPieces/useBoundPieces.js
usavkov-epam: Updated

mariia-aloshyna (approved):

alisher-epam (approved):

PR #2642, opened 2024-10-22
[UIIN-3097] "Something went wrong" screen when opening shared FOLIO/MARC instances in Inventory on member tenant

ryandberger:
- Fix [UIIN-3097](https://folio-org.atlassian.net/browse/UIIN-3097) by adding null safety on `currentPermission` object to prevent error screen.


zburke (approved):
Lint, this LGTM.

PR #2650, opened 2024-10-28
bump stripes-smart-components

Dmytro-Melnyshyn:
## Purpose
bump stripes-smart-components after https://github.com/folio-org/stripes-smart-components/pull/1532

### DIFF
--- a/package.json
+++ b/package.json
@@ -986,7 +986,7 @@
     "@folio/stripes-core": "^10.2.0",
     "@folio/stripes-inventory-components": "^1.0.0",
     "@folio/stripes-marc-components": "^1.0.1",
-    "@folio/stripes-smart-components": "^9.2.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -986,7 +986,7 @@
     "@folio/stripes-core": "^10.2.0",
     "@folio/stripes-inventory-components": "^1.0.0",
     "@folio/stripes-marc-components": "^1.0.1",
-    "@folio/stripes-smart-components": "^9.2.0",
+    "@folio/stripes-smart-components": "^10.0.0",
```
BogdanDenis, line 989: We can't just bump `stripes-smart-components` to a new version. Since we take it from `stripes` - it has to be a new stripes release with a new major version. Let's wait on Zak's comment in STSMACOM PR

PR #2661, opened 2024-11-01
UIIN-3115: ECS | "FOLIO/MARC-shared" source is displayed in manually shared instance record

mariia-aloshyna:
## Purpose

After instance sharing, the user gets an instance shadow copy instead of a shared instance. It happens because mod-search returned a response with `shared: true` but `tenantId: memberTenant` instead of the central tenant.

## Approach

- Check if instance is shared, if so - make the next request to mod-inventory with `x-okapi-tenant: centralTenant`

## Refs

https://folio-org.atlassian.net/browse/UIIN-3115

### DIFF
--- a/src/common/hooks/useInstance.js
+++ b/src/common/hooks/useInstance.js
@@ -3,18 +3,23 @@ import {
   useMemo,
 } from 'react';
 
+import { useStripes } from '@folio/stripes/core';
+
 import useSearchInstanceByIdQuery from './useSearchInstanceByIdQuery';
 import useInstanceQuery from './useInstanceQuery';
 
 const useInstance = (id) => {
+  const stripes = useStripes();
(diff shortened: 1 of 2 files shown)


src/common/hooks/useInstance.js
```diff
@@ -3,18 +3,23 @@ import {
   useMemo,
 } from 'react';
 
+import { useStripes } from '@folio/stripes-core';
```
BogdanDenis, line 6: Please import from `@folio/stripes/core` to use `stripes` peerDep

BogdanDenis (changes requested):

src/common/hooks/useInstance.js
mariia-aloshyna: automatic import got me there

Dmytro-Melnyshyn (approved):

PR #2681, opened 2024-11-29
UIIN-3147 User can edit Source consortium "Holdings sources" in member tenant but not in Consortia manager

NikitaSedyx:


usavkov-epam (approved):
Probably changelog should reflect this issue as well

SerhiiNosko (approved):

PR #2682, opened 2024-11-29
UIIN-3137: Add the `isSearchToggleHitInBrowse` flag to the history state to use in the condition that specifies whether to use the default sort on mount.

Dmytro-Melnyshyn:
## Purpose
Use previously applied sorting after switching from Browse search by pressing the Search toggle.

## Approach
Add a flag to `history.state` when hitting the Search toggle in Browse search, and use it to determine whether to use default sorting.

## Refs
[UIIN-3137](https://folio-org.atlassian.net/browse/UIIN-3137)

## Screenshots

https://github.com/user-attachments/assets/6ee6aad5-74ae-4556-9e79-4d9599612804


BogdanDenis (approved):

mariia-aloshyna (approved):

zburke (commented):
@Dmytro-Melnyshyn, this PR fixes one bug (losing state when toggling between search and browse) but introduces another in its place (losing state when toggling between this application and another). 

The screen-capture shows ui-inventory navigating from its default state [A], (sort: title) to a new state [B] (sort: contributor), out to another app, and then navigating back to ui-inventory. At this point, ui-inventory state should reflect [B], but it's [A]. This is incorrect, and not how other applications act. 

Please review the requested functionality with your PO before you merge. I'm happy to be involved in the discussion if there are questions about how application state should be preserved across navigation actions.

Dmytro-Melnyshyn:
@zburke, you are right, we should keep sorting between apps. I reverted this merge https://github.com/folio-org/ui-inventory/pull/2684

PR #2692, opened 2024-12-09
UIIN-3163 Add `limit=1000` to `subject-sources` request to fetch more subject sources

BogdanDenis:
## Description
By default okapi returns 10 items in a collection and that resulted Subject source facet to have some options to not display subject source names

## Screenshots

https://github.com/user-attachments/assets/6c922dcc-9db9-483d-ac20-3632c5b492e6

## Issues
[UIIN-3163](https://folio-org.atlassian.net/browse/UIIN-3163)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@
 * User can edit Source consortium "Holdings sources" in member tenant but not in Consortia manager. Refs UIIN-3147.
 * React 19: refactor away from react-dom/test-utils. Refs UIIN-2888.
 
+## [12.0.7] (IN PROGRESS)
+
+* Add `limit=1000` to `subject-sources` request to fetch more subject sources. Fixes UIIN-3163.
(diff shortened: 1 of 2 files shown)


Dmytro-Melnyshyn (approved):

mariia-aloshyna (approved):

zburke (approved):
This works of course, but note that `LIMIT` and `LIMIT_MAX` are sprinkled throughout this codebase already. Consider leveraging one of those existing constants instead of adding another hard-coded magic number.

BogdanDenis:
@zburke I see your point. Unfortunately, `LIMIT` and `LIMIT_MAX` don't work for us because one is `100` and one is `5000` 😄. I think it's ok to leave it as-is because adding `limit=1000` became such a common pattern, at least in Inventory and related modules, that it means "just give us more than 10 default records"

PR #2700, opened 2024-12-17
UIIN-3166: Remove the ability to share local instance when `Inventory: View, create instances` permission is assigned

OleksandrHladchenko1:
## Purpose
* Do not show **Share local instance** option when only `Inventory: View, create instances` permission is assigned. Share local instance action should only be available to users with the `Inventory: Share local instance with consortium` permission.

## Refs
[UIIN-3166](https://folio-org.atlassian.net/browse/UIIN-3166)


package.json
```diff
@@ -664,8 +664,7 @@
           "ui-inventory.instance.view",
           "inventory.instances.item.post",
           "inventory-storage.instances.item.post",
-          "browse.contributors.instances.collection.get",
-          "consortia.inventory.local.sharing-instances.execute"
+          "browse.contributors.instances.collection.get"
```
BogdanDenis, line 667: Is this a BE generated permission? I can't find it anywhere on the UI.
Also, there are 2 similar names: `consortia.inventory.local.sharing-instances.execute` which you've removed here. And there's `consortia.inventory.share.local.instance` that is used to define translations. Did we forget to update translations after changing permission names?

OleksandrHladchenko1: Yes, `consortia.inventory.local.sharing-instances.execute` is generated on [BE side](https://github.com/folio-org/mod-consortia/blob/81eceb708352b5e5fd41a98d2ecdfbfaa77a96db/descriptors/ModuleDescriptor-template.json#L496)

Also updated the translation path for **Inventory: Share local instance with consortium** permission

BogdanDenis (approved):

mariia-aloshyna (approved):

PR #2702, opened 2024-12-18
UIIN-3106: Add 'linked-data' interface to 'optionalOkapiInterfaces'

pkjacob:
## Purpose
Inventory UI makes API call to the following API in linked-data module
[GET linked-data/inventory-instance/:{id}/import-supported](https://github.com/folio-org/mod-linked-data/blob/master/descriptors/ModuleDescriptor-template.json#L95)

Per [stripes documentation](https://github.com/folio-org/stripes/blob/8790eaaccc2fa58f5500570f5bdc6a47c7a53ff3/doc/dev-guide.md), the optional interfaces that UI module uses should be declared in `optionalOkapiInterfaces` section of package.json file. 

Purpose of this PR is to update `optionalOkapiInterfaces` section of the package.json with the following
`"linked-data": "1.0"`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * Add call number browse settings. Refs UIIN-3116.
 * Remove the ability to share local instance when `Inventory: View, create instances` permission is assigned. Fixes UIIN-3166.
 * Display holdings names in `Consortial holdings` accordion for user without inventory permissions in member tenants. Fixes UIIN-3159
+* Add "linked-data 1.0" interface to "optionalOkapiInterfaces". Refs UIIN-3166.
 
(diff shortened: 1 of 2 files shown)


zburke (approved):
LGTM, though it's worth noting that the function leveraging this endpoint in [ViewInstance.js](https://github.com/folio-org/ui-inventory/blob/42c945284b28fae72300ab144f303a64253161e7/src/ViewInstance.js#L323-L344) is missing `<IfInterface>` protection, so it's nice that you're adding this declaration in `package.json` to include it in the module descriptor, but practically speaking the access is still unchecked and this is effectively a required interface.

BogdanDenis (changes requested):
Please add an entry in CHANGELOG

pkjacob:
@BogdanDenis  Change log added.

BogdanDenis (approved):

PR #2720, opened 2025-01-22
UIIN-3207: Sort holdings by location name and call number

julianladisch:
## Purpose
The instance view shows the holdings that belong to that instance. There are many libraries with many holdings per instance, they need this sorted by effectiveLocation.name, callNumberPrefix, callNumber, callNumberSuffix.

## Approach
Use this CQL query:

`instanceId==5b966ed6-3cd6-4dda-a0c7-f9b5e2826ccd sortBy effectiveLocation.name callNumberPrefix callNumber callNumberSuffix`

## Refs
https://folio-org.atlassian.net/browse/UIIN-3207

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 * Change import of `exportToCsv` from `stripes-util` to `stripes-components`. Refs UIIN-3025.
 * ECS: Disable opening item details if a user is not affiliated with item's member tenant. Fixes UIIN-3187.
 * Correctly depend on `inflected`. Refs UIIN-3203.
+* Decrease the amount of rerenders in `ConsortialHoldings` component. Fixes UIIN-3196.
 * Sort holdings by location name and call number. Refs UIIN-3207.
 
 ## [12.0.12](https://github.com/folio-org/ui-inventory/tree/v12.0.12) (2025-01-27)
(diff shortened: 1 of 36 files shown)


OleksandrHladchenko1 (approved):

BogdanDenis (approved):

zburke (approved):
1. OMG I cannot believe results have been unsorted all these years! Thank you for noticing this and fixing it. 
2. Please review the ticket with Ryan Taylor, UIIN PO, before merging. Given the display is "effective location - prefix - cn - suffix" it's hard to imagine he'll ask for a different order but for user-facing changes like this we should still ask.
3. But but but how does this work? Records returned by `/holdings-storage/holdings` don't include a property named `effectiveLocation.name` so how can you sort by it? 🪄 ✨ 🤨 and this property is not listed in [API documentation](https://s3.amazonaws.com/foliodocs/api/mod-inventory-storage/p/holdings-storage.html). How could a developer who is not familiar with the undocumented internal implementation of mod-inventory-storage come up with this solution?

zburke:
@julianladisch I appreciate the pointer to the documentation, but, ugh, strongly object the unusualness of the behavior in this API. I understand that it functions correctly and that is documented, but given the template-like nature of the documentation for fields like `offset`, `limit`, and `query` in RMB-based modules, I never looked at the documentation for query because I expected it to be the same as for every other module, and furthermore because you usually sort a result-set by a field present in the result-set, so I looked at the list of fields. 

Presumably you're doing the join (otherwise how could you sort by that field?) so just do the conventional thing and include `effectiveLocation { name }` in the result set. Then you don't even have to document it because it's just the normal, expected behavior.

mariia-aloshyna (approved):

PR #2733, opened 2025-02-05
UIIN-3205: Add `getCallNumberQuery` function to build a query string based on the call number and its type.

Dmytro-Melnyshyn:
## Description
In order to only return instances with the call number and call number type selected from call number browse settings, we need to use "Query search". When a user clicks on a call number title, they should be redirected to the Search page with a "Query search" option. It should have a call number query that takes into account the selected call number types.

The tests will pass after [PR](https://github.com/folio-org/stripes-inventory-components/pull/104) is merged, I didn't mock constants because tests will not catch the issue related to them if the constants become incorrect.

## Approach
Build a query in `getCallNumberQuery`. Take selected call number types from `useCallNumberBrowseConfig`.

## Refs
[UIIN-3205](https://folio-org.atlassian.net/browse/UIIN-3205)

## Screenshots

https://github.com/user-attachments/assets/3b89f0a9-bdb7-4db5-bbbb-7c5a3e55e4f7

### DIFF
--- a/src/components/BrowseResultsList/utils.js
+++ b/src/components/BrowseResultsList/utils.js
@@ -103,6 +103,14 @@ export const getFullCallNumber = (row) => {
   return fullCallNumber;
 };
 
+/**
+ * Constructs a query string for browsing call numbers based on the provided parameters.
+ *
+ * @param {string} qindex - The query index indicating the type of browse (e.g., callNumbers, dewey, lc).
+ * @param {Object} data - The data object containing configuration for call number browsing.
+ * @param {Object} row - The row object containing details of the item being browsed.
+ * @returns {string} - The constructed query string for browsing call numbers.
+ */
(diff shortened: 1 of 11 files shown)


BogdanDenis (approved):

src/components/BrowseResultsList/utils.js
```diff
@@ -103,22 +103,47 @@ export const getFullCallNumber = (row) => {
   return fullCallNumber;
 };
 
+const getCallNumberQuery = (qindex, data, row) => {
```
zburke, line 106: A function header that describes the arguments and return values sure would be nice.

zburke (approved):
LGTM, though consider adding some documentation around the new functions you are adding that describes their arguments and return values, especially those that generate complex CQL strings.

PR #2753, opened 2025-02-20
UIIN-3176: Instance: Suppress action menu and disable buttons when click Change log icon

OleksandrHladchenko1:
## Purpose
* On Instance view in Inventory, when you click the 'Version history' icon to open the Version history view in the 4th pane, the following should also occur in the 3rd pane:
- `Actions` menu button disappears
- `Tags` icon become greyed out and unclickable
- `Version history` icon become greyed out and unclickable

## Approach
* Hide/disable buttons when `isVersionHistoryOpen` is true.

## Refs
[UIIN-3176](https://folio-org.atlassian.net/browse/UIIN-3176)

## Screenshots
https://github.com/user-attachments/assets/7d2d1c01-c24f-4f11-81f0-1edbf761d799

### DIFF
--- a/src/Instance/InstanceDetails/InstanceDetails.test.js
+++ b/src/Instance/InstanceDetails/InstanceDetails.test.js
@@ -57,7 +57,7 @@ const mockReferenceData = {
 
 const queryClient = new QueryClient();
 
-const actionMenu = jest.fn();
+const actionMenu = jest.fn(() => <span>Action menu</span>);
 const onClose = jest.fn();
 const tagsEnabled = true;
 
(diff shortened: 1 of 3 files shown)


BogdanDenis (approved):

Dmytro-Melnyshyn (approved):

src/Instance/InstanceDetails/InstanceDetails.test.js
```diff
@@ -57,7 +57,7 @@ const mockReferenceData = {
 
 const queryClient = new QueryClient();
 
-const actionMenu = jest.fn();
+const actionMenu = jest.fn(() => <span>Action menu</span>);
```
mariia-aloshyna, line 60: semantically it would be better to use a button tag instead of a span

PR #2765, opened 2025-02-27
UIIN-3257 include limit param in jobProfiles query

zburke:
Without a `limit` clause, CQL will only return 10 records (the default) even when more than 10 identifiers are specifically passed in the query. This suprises me, but apparently it is by design.

Follows up on PR #1820 where we added a `limit` clause to retrieve all providers; this PR makes sure we retrieve all a provider's job profiles. 

Refs [UIIN-3257](https://folio-org.atlassian.net/browse/UIIN-3257)

### DIFF
--- a/src/components/ImportRecordModal/ImportRecordModal.js
+++ b/src/components/ImportRecordModal/ImportRecordModal.js
@@ -14,9 +14,8 @@ import { Loading, Modal, Select, TextField, ModalFooter, Button } from '@folio/s
 import { buildQueryByIds } from '../../utils';
 
 // upper limit of how many job profiles may be retrieved for a given target.
-// a limit clause is necessary because a query like 
-// `query=id==A or id==B or ... Z` will only retrieve 10 (the default limit) 
-// even when more than 10 ids are specified.
+// a limit clause is required to return anything other than the default number
+// of records.
 //
 // 60 is not an arbitrary number; it's a back-of-the-envelope calculation
 // intended to make sure we stay under the query-string's length-limit.
--- a/src/components/ImportRecordModal/ImportRecordModal.js
+++ b/src/components/ImportRecordModal/ImportRecordModal.js
@@ -13,6 +13,16 @@ import { Loading, Modal, Select, TextField, ModalFooter, Button } from '@folio/s
 
 import { buildQueryByIds } from '../../utils';
 
+// upper limit of how many job profiles may be retrieved for a given target.
+// a limit clause is necessary because a query like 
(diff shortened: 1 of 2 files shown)


src/components/ImportRecordModal/ImportRecordModal.js
```diff
@@ -13,6 +13,16 @@ import { Loading, Modal, Select, TextField, ModalFooter, Button } from '@folio/s
 
 import { buildQueryByIds } from '../../utils';
 
+// upper limit of how many job profiles may be retrieved for a given target.
+// a limit clause is necessary because a query like 
+// `query=id==A or id==B or ... Z` will only retrieve 10 (the default limit) 
+// even when more than 10 ids are specified.
```
MikeTaylor, line 19: This comment is a bit misleading. It's not that the query retrieves 10 records. It doesn't retrieve _any_ records. It's that the query merely specifies a set of records that we're interested in knowing about; and the separate `limit` parameter says how many of them we actually want to see.

MikeTaylor (approved):
Looks good, though I remain mildly unhappy about the wording of the comment.

mariia-aloshyna (approved):
thank you for the PR, there are a couple of eslint errors

src/components/ImportRecordModal/ImportRecordModal.js
zburke: I sort of see your point, but I see a difference in asking for known vs. unknown quantities. If I ask, "Could you get me everything matching foo?" then I accept "There could be many, we'll give you 10 by default" a reasonable response. But if I say, "Could you get me everything on this list of 11 things?" then it seems petty to say, "We'll give you 10 by default unless you tell us how many you want."

But knowing that this is how it works and that whining isn't going to change it, I have tried to provide a less snarky comment.

MikeTaylor: Your new comment is indeed less petty :-)

I know we rehearsed this in Slack already, but the issues are twofold:
1. Philosophically, the question of which set of records you're interested in and how many you want returned to you are quite separate.
2. If you want to special-case a certain set of queries, you need to define which ones should be included:
  * `id=123`
  * `id=123 or id=124`
  * `id=(123 or 124)`
  * `id=123 or 124`
  * `id=(123 and 124) and author=burke`
  * `hrid=in0001`
  * `hrid=in0001 or id=123`
  * _etc_.

Anyway, we are certainly sufficiently in agreement for this PR to be merged.

PR #2766, opened 2025-02-27
UIIN-3228: Use the new call number browse functionality only if 1.5 `browse` interface is available.

Dmytro-Melnyshyn:
## Description
Call Number Browse refactor changes need to be backported to Ramsons release.
Since the refactor includes breaking changes (new browse interface version) we need to add checks in new code to see which version of `browse` interface is available.
If the old version is available - we should default to old logic.
And if the new version is available - we can safely use new logic for browsing.

Changes that don’t require such checks:
- new settings for call number type sorting (https://folio-org.atlassian.net/browse/UIIN-3116 )

Changes that require checks:
- Call Number Browse | Refactor to use new browse endpoint (https://folio-org.atlassian.net/browse/UIIN-3162 )
- CLONE -Call Number Browse | Refactor to use new browse endpoint (https://folio-org.atlassian.net/browse/UISINVCOMP-48 )
- Use query search when navigating from call number browse to search (https://folio-org.atlassian.net/browse/UIIN-3205 )
- Remove hover-over text next to "Shelving order" on the Item record detail view (https://folio-org.atlassian.net/browse/UIIN-3210 )

Tests will pass after the https://github.com/folio-org/stripes-inventory-components/pull/113 is merged.

## Related PRs
https://github.com/folio-org/stripes-inventory-components/pull/113

## Refs
[UIIN-3228](https://folio-org.atlassian.net/browse/UIIN-3228)


package.json
```diff
@@ -54,7 +54,7 @@
     "okapiInterfaces": {
       "alternative-title-types": "1.0",
       "call-number-types": "1.0",
-      "browse": "0.6 1.0",
+      "browse": "0.6 1.0 1.5",
```
BogdanDenis, line 57: Let's add this to `optionalOkapiInterfaces` instead. This way it won't mean a breaking change

src/components/BrowseResultsList/BrowseResultsList.js
```diff
@@ -65,20 +69,20 @@ const BrowseResultsList = ({
   const listId = `browse-results-list-${browseOption}`;
 
   const isSelected = useCallback(({ item, rowIndex }) => {
-    if (isRowPreventsClick(item, browseOption)) return false;
+    if (isRowPreventsClick(item, browseOption, stripes)) return false;
```
BogdanDenis, line 72: These utility functions don't really need the whole stripes object, we can just pass a boolean value like `isNewCallNumberBrowseAvailable`
Same goes for `getBrowseResultsFormatter`

src/filterConfig.js
```diff
@@ -23,3 +23,29 @@ export const browseInstanceIndexes = [
   { label: 'ui-inventory.browse.contributors', value: `${browseModeOptions.CONTRIBUTORS}`, queryTemplate: '%{query.query}' },
   { label: 'ui-inventory.browse.subjects', value: `${browseModeOptions.SUBJECTS}`, queryTemplate: '%{query.query}' },
 ];
+
+export const _browseInstanceIndexes = [
```
BogdanDenis, line 27: Let's call this `newCallNumberBrowseInstanceIndexes` and add a comment to explain why we have 2 sets of indexes.

src/hooks/useInventoryBrowse/constants.js
```diff
@@ -22,6 +23,17 @@ export const PATH_MAP = {
   [browseModeOptions.CONTRIBUTORS]: 'browse/contributors/instances',
 };
 
+export const _PATH_MAP = {
```
BogdanDenis, line 26: Same about indexes here - we should add some explanation why there are two almost identical constants

BogdanDenis (changes requested):

BogdanDenis (approved):

PR #2768, opened 2025-02-27
UIIN-3260: Update Inventory: Set records for deletion permission's effect on ‘Set for deletion’ checkbox in Instance Edit view

mariia-aloshyna:
## Purpose

Users without this permission will be able to see but NOT edit the 'Set for deletion' checkbox in Instance Edit view

The ‘Set for deletion’ checkbox should be greyed out and uneditable for users who don’t have the necessary permission.

## Approach

- Grey out Set for deletion field if no permission.

## Refs

https://folio-org.atlassian.net/browse/UIIN-3260


OleksandrHladchenko1 (approved):
Fix eslint error in `src/edit/InstanceForm.js` please

PR #2782, opened 2025-03-10
UIIN-3251: Use `stripes.okapi.tenant` in order to get persistent id between pages

OleksandrHladchenko1:
## Purpose
* When cancel moving holdings/items to another instance, buttons to move holdings/items disappear.

## Approach
* After cancellation moving `holdings-storage/holdings` endpoint uses central `tenantId` instead of current tenant id.

## Refs
[UIIN-3251](https://folio-org.atlassian.net/browse/UIIN-3251)


zburke (approved):
There are _so many_ different tenantId values floating around in this component 😢 
* instance.tenantId
* stripes.okapi.tenant
* stripes.user.user?.consortium?.centralTenantId
* stripes.user.user?.consortium?.id

It would be helpful to have comments describing why we use one or the other in each different place, e.g. why is `useInstanceMutation` able to use a value from the instance or from stripes, but `<ViewInstance>` always has to get it from stripes in order to resolve this bug.

PR #2786, opened 2025-03-12
UIIN-3261: ViewSource - pass `tenantId` to `MarcVersionHistory`, disable Actions dropdown.

Dmytro-Melnyshyn:
## Description
1. Pass `tenantId` to `MarcVersionHistory` to retrieve settings from the correct tenant. 
2. Disable the `Actions` dropdown in `ViewSource` when the record history pane is open. It requires using the `Dropdown` component because the `actions` passed to the `Pane` cannot be disabled.

## Refs
https://folio-org.atlassian.net/browse/UIIN-3261

## Screenshots
image


zburke (approved):
Please include a CHANGELOG entry

Dmytro-Melnyshyn:
> Please include a CHANGELOG entry

This is a follow-up PR, I forgot to mention that.

PR #2794, opened 2025-03-18
UIIN-3269: Disable hyperlink when user doesn't have `users.item.get` permission (follow-up)

OleksandrHladchenko1:
## Approach
* Add `users.item.get` to all permissions in order to see user with all inventory permissions. User will see the hyperlink only when "Users: Can view user profile" permission is assigned

## Refs
[UIIN-3269](https://folio-org.atlassian.net/browse/UIIN-3269)


BogdanDenis (approved):

mariia-aloshyna (approved):

OleksandrHladchenko1:
> This change is very specific, only enabling links for folks with the "all" permission set. Even somebody with `ui-inventory.instance.view` permission (where `users.collection.get` is included but `users.item.get` is not) won't see a link in the version history. Are you confident this is correct or are there other psets that should include `users.item.get`? Unfortunately, the jira story is not detailed enough to assess this.

Specific requirements require specific changes. The requirement was to display user name as plain text when user have only `ui-inventory.instance.view`. `users.collection.get` is needed to get the list of the users. Without this permission the request to get users fails. So, we need to rely on `users.item.get` whether display link or not.

PR #2814, opened 2025-04-07
UIIN-3122 wait for central tenant permissions to load before focusing third pane (follow-up)

BogdanDenis:
## Description
Need to wait for central tenant permissions to load because it is also a condition to show loading pane. Focusing before permissions are loaded has no effect on the third pane.

## Issues
[UIIN-3122](https://folio-org.atlassian.net/browse/UIIN-3122)

### DIFF
--- a/src/ViewInstance.js
+++ b/src/ViewInstance.js
@@ -244,6 +244,7 @@ class ViewInstance extends React.Component {
     const {
       selectedInstance,
       stripes,
+      isCentralTenantPermissionsLoading,
     } = this.props;
     const isMARCSourceRecord = isMARCSource(selectedInstance?.source);
     const isLinkedDataSourceRecord = isLinkedDataSource(selectedInstance?.source);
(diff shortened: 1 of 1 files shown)


UladzislauKutarkin (approved):

zburke (approved):
Code change LGTM but ... should there be any test changes? CHANGELOG entry?

BogdanDenis:
> Code change LGTM but ... should there be any test changes? CHANGELOG entry?

This is a follow-up PR to a previous PR, so there is a CHANGELOG entry for this ticket already.
As for tests, I don't think Jest/RTL support checking if an element has focus, since it's not a real DOM environment, but jsdom.

zburke:
@BogdanDenis, WRT focus: [we use jest-dom](https://github.com/folio-org/jest-config-stripes/blob/d0a637f23812cf21b1d776f82b9917a476facf40/package.json#L23) and it supports [focus assertions](https://github.com/testing-library/jest-dom?tab=readme-ov-file#tohavefocus).

PR #2822, opened 2025-04-23
UIIN-3211: Remove extra record from Central tenant after adding Receiving history record in local member instance

OleksandrHladchenko1:
## Purpose
* After adding `Receiving history` record in local member instance, `Receiving history` accordion should contain one record in Member tenant (as expected). Currently it contains two identical records - one in Member tenant (as expected) and extra record in Central tenant.

## Approach
* Updated `useReceivingHistory` hook. Central tenant cannot create version history by editing holding. So the user gets pieces only from API.

## Refs
[UIIN-3211](https://folio-org.atlassian.net/browse/UIIN-3211)

## Screenshots
https://github.com/user-attachments/assets/485c3778-9d92-4674-ae89-d7f46f3a8c90


src/Holding/ViewHolding/HoldingReceivingHistory/useReceivingHistory.js
```diff
@@ -7,9 +7,9 @@ import {
 } from '@folio/stripes/core';
 import { LIMIT_MAX } from '@folio/stripes-inventory-components';
 
-
 const useReceivingHistory = (holding, tenant, options = {}) => {
   const stripes = useStripes();
+  const centralTenantId = stripes.user.user.consortium.centralTenantId;
```
mariia-aloshyna, line 12:
```suggestion
  const centralTenantId = stripes.user.user?.consortium?.centralTenantId;
```
lets make it optional to avoid errors when it is not consortium mode

Dmytro-Melnyshyn (approved):

mariia-aloshyna (approved):

PR #2829, opened 2025-05-27
UIIN-3366 Use `/search/resources/jobs` endpoints to get Instances and Holdings UUIDs

BogdanDenis:
## Description
Old `/search/instances/ids` and `/search/holdings/ids` endpoints are deprecated, and we have to use new and better ones that work via jobs.
New workflow is:
1. Create a job to export ids, provide a search query and a record type
2. Wait until the job completes
3. Make a request to `/search/resources/jobs/{id}/ids` to get a list of ids

If a job takes a long time to complete - we'll be polling it (default timeout is 5s) until the status is not `IN_PROGRESS`. If it fails - we'll show a generic error message just as before.

This new logic is placed in a hook, but unfortunately `<InstancesList>` component is still class-based and uses a mutator to make request. This means that a HOC is needed to inject `useResourcesIds` values into the component.
It's a bit of a fuss, but doing this vs having to mess with mutators to do this I think HOCs are a bit more convenient. Plus it means less refactoring when we'll eventually make `<InstancesList>` a functional component.
Created a directory for HOCs, and moved existing HOCs to all be in one place.

## Screenshots

https://github.com/user-attachments/assets/75818091-0209-4292-9f83-42cc852b388a

## Issues
[UIIN-3366](https://folio-org.atlassian.net/browse/UIIN-3366)


zburke (approved):
1. 👏 👏 👏 This is nice work, well-explained. It meets the immediate needs of this ticket and improves the long-term usability of this package. Thank you, thank you!
2. It's nice to drop the GET query-batching in favor of the POST API when handling long lists. So much simpler. I know that's not even your code, but still, it's always great when we can remove work-arounds from our codebase. 
3. At risk of bikeshedding, why use `ky` instead of [react-query and its `{ refetchInterval }` query-option](https://tanstack.com/query/v3/docs/framework/react/reference/useQuery)? Given you already have working code here, I am not suggesting that you do this now, but rather that you keep react-query front-of-mind for future work. Of course ky is capable of doing this job, as you have proved, but react-query handles a lot of that same work automatically.

src/hooks/useResourcesIds/useResourcesIds.test.js
```diff
@@ -0,0 +1,139 @@
+import {
+  renderHook,
+  act,
```
mariia-aloshyna, line 3: remove unused import please

mariia-aloshyna (approved):

PR #2856, opened 2025-07-18
UIIN-3445: Fix "New request" button is missing from the "Actions" dropdown in the Instance details pane

mariia-aloshyna:
## Purpose

"New request" button is missing from the "Actions" dropdown in the Instance details pane.

## Approach

- Add hook to fetch TLR settings when opening instance details.
- If TLR = 'true' - show 'New request' action item.

## Refs

https://folio-org.atlassian.net/browse/UIIN-3445


src/hooks/useTLRSettingsQuery/useTLRSettingsQuery.js
```diff
@@ -0,0 +1,18 @@
...
+  useNamespace,
+  useOkapiKy,
+} from '@folio/stripes/core';
+
+const useTLRSettingsQuery = (tenant) => {
+  const ky = useOkapiKy({ tenant });
+  const [namespace] = useNamespace({ key: 'tlr-settings' });
+
+  return useQuery(
+    [namespace],
```
Dmytro-Melnyshyn, line 13:
```suggestion
    [namespace, tenant],
```

PR #2865, opened 2025-07-28
UIIN-3450: Fix instance overlaying issues

OleksandrHladchenko1:
## Purpose
* After overlaying the instance green success message does not appear. The instance is not updated. The green success message should appear. The updated instance is displayed in the 3rd pane.

## Approach
* Pass args to API call.
* Update endpoint to overlay instance.

## Refs
[UIIN-3450](https://folio-org.atlassian.net/browse/UIIN-3450)

## Screenshots
https://github.com/user-attachments/assets/ef29156a-357c-4d19-9520-cb06cb38c19f

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ UIIN-3437.
 * Open Item record detail view when exact match is found via Keyword search option in Item tab. Refs UIIN-3416.
 * Classification browse | Add Title to results list. Refs UIIN-3367.
 * CallNumberTypes component: refactor class component. Refs UIIN-3389.
+* Fix instance overlaying issues. Fixes UIIN-3450.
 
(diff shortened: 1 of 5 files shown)


zburke (approved):
Wow, there are so many changes in this PR, it's hard to understand how the code got into this badly broken state. It would be nice to have a little more history/context for what happened here so we can understand _why_ this change is necessary. Are we really fixing bugs, or is this a migration from one approach (import via `/inventory/import/$id`) to another (import via `copycat/imports`)? If possible, try to summarize the approach in the PR title/commit subject: "fix issues" is not very specific, e.g. "Migrate instance-overlay to copycat pipeline" (if that's actually what is happening here). 

Consider adding a test to ViewInstance that checks for the success modal since that is clearly the expectation from the ticket but that feature was blatantly unimplemented in the code. 

Please address the lint errors before you merge.

PR #2868, opened 2025-07-31
UIIN-3453: Display order column in items table when clicking `Move items within an instance` action button

OleksandrHladchenko1:
## Purpose
* Add a new “Order“ column to the items list when re-ordering items within an instance.
* When observing an instance details pane with holdings and items in it
* Populate new column in items list named 'Order' after a column with 'drag&drop' icon and before checkboxes column
* Item’s order should be retrieved from BE
* When “Move items within an instance” action is invoked, sorting should be disabled.
* When clicking ‘Stop items movement within an instance' action item, enable sorting

## Approach
* When `draggable` is true then disable sorting and add new column "Order"
* Increase test coverage

## Refs
[UIIN-3453](https://folio-org.atlassian.net/browse/UIIN-3453)

## Screenshots
https://github.com/user-attachments/assets/95c148b9-4946-4889-9337-c39f32216bd5

### DIFF
--- a/src/Instance/ViewInstance/ViewInstance.js
+++ b/src/Instance/ViewInstance/ViewInstance.js
@@ -293,7 +293,7 @@ const ViewInstance = (props) => {
       refetch();
       callout.sendCallout({
         type: 'success',
-        message: <FormattedMessage id="ui-inventory.copycat.callout.updated" values={{ xid: instance?.id }} />,
+        message: <FormattedMessage id="ui-inventory.copycat.callout.updated" values={{ xid: args.externalIdentifier }} />,
       });
     };
 


src/Instance/ItemsList/ItemsList.js
```diff
@@ -247,7 +250,7 @@ const ItemsList = ({
       columnWidths={columnWidths}
       pagingType={MCLPagingTypes.PREV_NEXT}
       totalCount={total}
-      nonInteractiveHeaders={['loanType', 'effectiveLocation', 'materialType']}
+      nonInteractiveHeaders={draggable ? visibleColumns : ['order', 'loanType', 'effectiveLocation', 'materialType']}
```
mariia-aloshyna, line 253: shouldn't be order sortable when non dragging?

translations/ui-inventory/en_US.json
```diff
@@ -309,6 +309,7 @@
     "fetchingMarcRecord": "Fetching MARC record",
     "viewMarcSource": "View MARC source",
     "item.barcode": "Item: barcode",
+    "item.order": "Order",
```
mariia-aloshyna, line 312: there is no need to change other from `en.json` files

src/Instance/ItemsList/ItemsList.js
OleksandrHladchenko1: Added sorting by Order column

PR #2869, opened 2025-07-31
UIIN-3422 Allow tenant to set default columns to display in Inventory results

BogdanDenis:
## Description
Added "Default columns" checkbox group to "Display settings". This setting will control which columns are visible by default in Instances search results list. User can still personalize which columns they can see via "Actions" menu in Instances search, but this selection will be reset when closing the tab (as before).

## Screenshots

https://github.com/user-attachments/assets/acf37b9b-1f15-431a-8be5-ee3bf5419067

## Issues
[UIIN-3422](https://folio-org.atlassian.net/browse/UIIN-3422)


mariia-aloshyna (commented):
fix eslint issues, please

mariia-aloshyna (approved):

PR #2870, opened 2025-08-05
UIIN-3328-improve-alignment-and-hierarchy-within-holdings-and-item-forms

mhamann-ubl:
## Purpose
After reviewing the changes made with PRs regarding UIIN-3327 and UIIN-3328 there were some suggestions from the POs involved on improving the UI. This PR addresses these requests.

## Approach
It improves alignment within the new additional call number component, but also alignment between different components. It makes use of the Headline Element instead of the preexisting use of <strong> Attributes, to introduce a proper hierarchy between elements.

## Refs
https://folio-org.atlassian.net/browse/UIIN-3328
https://folio-org.atlassian.net/browse/UIIN-3327


mhamann-ubl:
> Hi @mhamann-ubl it looks better now, but `improve-ui` still says nothing, each PR aims to improve UI. Be more specific, what changes have been introduced?

Hi @mariia-aloshyna it left it like this, because it refers to the PRs that we have already merged: UIIN-3327 and 3328. It improves upon the UI within the range of these features. These are described with the original PRs. Does that make sense?

mariia-aloshyna (approved):

PR #2893, opened 2025-09-10
UIIN-3471: Add default values

elsenhans:
https://folio-org.atlassian.net/browse/UIIN-3471

## Purpose
Correcting the behavior of save button, drop-down menu and checkbox in Settings > Inventory > Number generator options by adding default values for the form.
The save button will not be enabled if the initial default values are selected.

See also first changes of this issue:
https://github.com/folio-org/ui-inventory/pull/2877


BogdanDenis (approved):
Please add an entry in CHANGELOG for this ticket

Dmitriy-Litvinenko (approved):

PR #2904, opened 2025-09-26
UIIN-3486: Show order cells as text inputs when invoking “Move items within an instance” action

mariia-aloshyna:
## Purpose

Allow for libraries to change the numeric sort order of Items via Inventory UI manually by invoking the existing “Move items within an instance” action and save their selections by invoking the existing “Stop items movement within an instance” action.

## Approach

- Create hook that provides a comprehensive solution for managing item order changes in the inventory system. It handles validation, auto-adjustment, state management, and API integration for reordering items within a holding.
- Display order cell as number input
- Add validation
- Update item orders when invoking "Stop items movement" action

## Refs

https://folio-org.atlassian.net/browse/UIIN-3486

## Screenshots

https://github.com/user-attachments/assets/6de01cf1-68e8-4543-8a4a-bcd4081545ec


BogdanDenis (approved):
Implementation lgtm, my only concern is with the width of the inputs. I believe there can be tens and even hundreds of Items in a Holdings record. Will the user be able to see the order number when it has 2 or 3 digits?

zburke (approved):
Wow, tons of great work here. I appreciate all the care that went into creating a thoughtful API with the new context and hooks. The comments, though few, are very helpful. Thank you!

I share @BogdanDenis's concern about the narrow, default width of the input field. Already the field is impossible to read when it contains invalid input and gets the ❌ overlay, as can be seen in the recording at 00:33. Can we shrink the first column which is nothing but anchors, or the checkbox column, or "Copy number" (make it "Copy #"?!?) to buy ourselves a few pixels?

PR #3006, opened 2026-04-21
UIIN-3605: Holdings-record detail path with invalid UUIDs shows never-ending loading spinner

mariia-aloshyna:
## Purpose

Accessing a URL like `/inventory/view/foo/bar` shows two stripes-connect error alerts, indicating unhanded API request failures, followed by a never-ending loading spinner.

## Approach

- Handle API failures and show error callout with message from the response or default communication error.
- Check for valid holdings and instances UUIDs, if not valid - show default `404 Error` page from `stripes-core` (needs to be directly exported from stripes-core).
- Add unit tests

Merge after https://github.com/folio-org/stripes-core/pull/1708

## Refs

https://folio-org.atlassian.net/browse/UIIN-3605

## Screenshots

Screenshot 2026-04-21 at 13 05 57


src/routes/ViewHoldingRoute.js
```diff
@@ -3,39 +3,57 @@ import {
   useParams,
   useLocation,
 } from 'react-router-dom';
+import { FormattedMessage } from 'react-intl';
 
 import { useStripes } from '@folio/stripes/core';
+import { AuthenticatedError } from '@folio/stripes-core/src/components';
```
BogdanDenis: Looks like the import path can be updated now when the stripes-core PR is merged

src/routes/ViewHoldingRoute.js
```diff
@@ -3,39 +3,57 @@ import {
...
+  const callout = useCallout();
+  const location = useLocation();
+  const { instance, error } = useSearchInstanceByIdQuery(instanceId);
   const { updateOwnership } = useUpdateOwnershipMutation(UPDATE_OWNERSHIP_API.HOLDINGS);
 
   const { settings } = useAuditSettings({ group: INVENTORY_AUDIT_GROUP });
   const isVersionHistoryEnabled = getIsVersionHistoryEnabled(settings);
 
+  if (!UUID_PATTERN.test(instanceId) || !UUID_PATTERN.test(holdingsrecordid)) {
+    return <AuthenticatedError location={location} />;
```
BogdanDenis, line 43: I think it's possible to use regexp in react-router's `path` prop - if that's the case then we can check for valid uuids there and also add a generic route for not valid routes, which would handle all other possible cases of not valid paths

src/ViewHoldingsRecord.js
```diff
@@ -663,6 +681,10 @@ class ViewHoldingsRecord extends React.Component {
       isVersionHistoryOpen,
     } = this.state;
 
+    if (this.isHoldingNotFound()) {
```
BogdanDenis: I think going to `/inventory/instances/foo/bar` and a record not being found are two different cases.
It's great to handle the first one. And for the second I believe currently we're returning the user to a previous page and showing a callout message? If we change how we handle not found records it could potentially also break a lot of e2e tests

src/routes/ViewHoldingRoute.js
mariia-aloshyna: do you have some examples of validating uuids directly in `path` prop? and how to handle invalid uuids then? in ui-inventory there is `react-router v5.0.1` and I haven't found any info in doc about route validating except component-level validation.

OleksandrHladchenko1 (approved):

src/routes/ViewHoldingRoute.js
```diff
@@ -3,39 +3,59 @@ import {
...
+  }
+
+  if (error) {
+    const defaultErrorMessage = <FormattedMessage id="ui-inventory.communicationProblem" />;
+
+    callout.sendCallout({
+      type: 'error',
+      message: error?.message || defaultErrorMessage,
+    });
```
zburke, line 53: Set `timeout: 0` so the callout is manually (instead of automatically) dismissed. This is our typical pattern when showing error callouts. I'm trying to find documentation for this pattern, or better yet, getting Callout updated so `type: "error"` uses `timeout: 0` automatically.

zburke (changes requested):
This change works better than the master branch, but I don't feel like it meets the "expected behavior" described in the ticket. I have verified that my stripes-core install includes STCOR-1047. 

1. A URL with correctly-shaped but invalid/nonexistent UUIDs now shows a pair of "Not found" error toasts that disappear, reappear, disappear. There is not a permanent 404-type message in the application pane; once the callouts disappear, the screen is blank except for the navbar.
2. A URL with incorrectly-shaped (i.e. completely invalid) UUIDs shows nothing, just a blank screen except for the nav bar. There is no callout, no error message.

Aside, do you think we should export a UUID-validation function from stripes-util? There's nothing "inventory-ish" about this function and UUIDs are everywhere in FOLIO. On the one hand, eh, NBD, OTOH, the regex is not completely trivial. Dunno. Just thinkin'.

src/routes/ViewHoldingRoute.js
BogdanDenis: v5 does support it, and we can define route paths like `/view/:id([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})`.
Stated here, but not very clearly - https://v5.reactrouter.com/web/api/Route/path-string-string
But this feature was removed in `react-router v6`, and I wasn't aware of this. So if there are plans to migrate to `react-router v6` let's keep validation inside routes.

mariia-aloshyna: it doesn't look safely TBH, in case if one day we decide to migrate to `react-router v6` I'm not sure someone would recall that this feature was removed

src/routes/ViewHoldingRoute.js
mariia-aloshyna: Hi @zburke attaching screen recording with scenario 1 (correctly-shaped but invalid UUIDs). You sure that you are on a branch UIIN-3605?
I see that `ui-users` module returns just `User not found` message on the details page, do you think it is better way to handle such scenario? Or it is better to figure out with PO and establish common pattern for such cases?
Regarding moving regex to stripes-util I think it is a good idea, I will prepare a PR.

https://github.com/user-attachments/assets/2083a26f-5634-41c3-9c16-8b7a75743b9d

BogdanDenis (approved):

PR #3008, opened 2026-04-21
UIIN-3602: Fix Tags accordion displaying in Holdings and Items until `Tags` setting is re-enabled

OleksandrHladchenko1:
## Purpose
* When the Tags setting is enabled, Tags should be displayed on each Folio record when applicable.

## Approach
* Empty settings means tags are enabled by default

## Refs
[UIIN-3602](https://folio-org.atlassian.net/browse/UIIN-3602)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change history for ui-inventory
 
+## [14.0.1] (IN PROGRESS)
+
+* Fix Tags accordion displaying in Holdings and Items until Tags setting is re-enabled. Fixes UIIN-3602.
+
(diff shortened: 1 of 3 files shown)


zburke (approved):
I look forward to that glorious future day when stripes-connect resources have been fully replaced by react-query. Until then, this looks good.

mariia-aloshyna (approved):

mariia-aloshyna:
> I look forward to that glorious future day when stripes-connect resources have been fully replaced by react-query. Until then, this looks good.

Hi @zburke `ViewItem` already uses `react-query` to fetch `tagSettings`, as for `ViewHoldings` we have a ticket to refactor the component to be functional one instead of class, and in scope of it we plan to replace stripes-connect resources :face_exhaling:

PR #3011, opened 2026-04-24
UIIN-3635 `<ViewSource>` - don't load MARC record until Instance has finished loading.

BogdanDenis:
## Description
On a member tenant, when a View Source page is refreshed (or opened via copy-pasting a URL or duplicating a tab) we get an error that a MARC record was not found. This is caused by fetching MARC source record before Instance has finished loading.
We need that Instance record to know it's tenantId to know which tenant to request a MARC source record from

## Issues
[UIIN-3635](https://folio-org.atlassian.net/browse/UIIN-3635)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [14.0.1] (IN PROGRESS)
 
 * Fix Tags accordion displaying in Holdings and Items until Tags setting is re-enabled. Fixes UIIN-3602.
+* `<ViewSource>` - don't load MARC record until Instance has finished loading. Fixes UIIN-3635.
 
 ## [14.0.0](https://github.com/folio-org/ui-inventory/tree/v14.0.0) (2026-04-17)
 [Full Changelog](https://github.com/folio-org/ui-inventory/compare/v13.0.13...v14.0.0)
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@folio/inventory",
-  "version": "14.0.0",
+  "version": "14.0.1",
   "description": "Inventory manager",
   "repository": "folio-org/ui-inventory",
   "publishConfig": {
--- a/src/components/ViewSource/ViewSource.js
+++ b/src/components/ViewSource/ViewSource.js
@@ -145,6 +145,10 @@ const ViewSource = ({
   ], [stripes, redirectToMARCEdit]);
 
   useEffect(() => {
+    if (isInstanceLoading) {
+      return;
+    }
+
     setIsMarcLoading(true);
 
     const { okapi: { tenant, token, locale } } = stripes;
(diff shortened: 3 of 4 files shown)


mariia-aloshyna (approved):

zburke (commented):
The `useEffect()` implementation pulls the tenant from props (when present) or by default from stripes. This doesn't quite line up with the explanation given in the PR description, which describes deriving the tenant from the instance record. If this fix works, however, that suggests `<ViewSource>` is maybe being called like this:
```
const { instance, isInstanceLoading } = useGetInstance();
return <ViewSource 
  instance={instance}
  isInstanceLoading={isInstanceLoading}
  tenantId={instance.tenantId}
  ... 
  />
```
which is fine, but it suggests the current `useEffect` implementation is buggy: instead of allowing `props.tenantId` to fallback to `stripes.okapi.tenant` (maybe this is the true source of this bug), `<ViewSource>` should exclusively use `props.tenantId`. 

Does this make sense? Or is it only true that `props.tenantId` is required in an ECS env, whereas in single-tenant mode we can fallback onto `stripes.okapi.tenant` because that's what `props.tenantId` will eventually fulfill to anyway.

OleksandrHladchenko1 (approved):

BogdanDenis:
@zburke 
> The `useEffect()` implementation pulls the tenant from props (when present) or by default from stripes. This doesn't quite line up with the explanation given in the PR description, which describes deriving the tenant from the instance record. If this fix works, however, that suggests `<ViewSource>` is maybe being called like this:
> 
> ```
> const { instance, isInstanceLoading } = useGetInstance();
> return <ViewSource 
>   instance={instance}
>   isInstanceLoading={isInstanceLoading}
>   tenantId={instance.tenantId}
>   ... 
>   />
> ```

Yes, that's how `<ViewSource>` is used - `props.tenantId` is the instance's tenant id, and so we need to wait until the instance is loaded to request MARC source record with the correct tenant id.

> which is fine, but it suggests the current `useEffect` implementation is buggy: instead of allowing `props.tenantId` to fallback to `stripes.okapi.tenant` (maybe this is the true source of this bug), `<ViewSource>` should exclusively use `props.tenantId`.
> 
> Does this make sense? Or is it only true that `props.tenantId` is required in an ECS env, whereas in single-tenant mode we can fallback onto `stripes.okapi.tenant` because that's what `props.tenantId` will eventually fulfill to anyway.

I think it makes sense to remove the fallback to `stripes.okapi.tenant`. If an instance record doesn't contain `tenantId` for some reason - then we'll be sending a request with the wrong tenantId anyways.

zburke (approved):
> I think it makes sense to remove the fallback to `stripes.okapi.tenant`. 

Explaining this to myself to make sure I understand it: 

This fixes a bug in ECS envs (where we needed to pull the tenant off the instance record because it might not match the one on `stripes`) at the expense of a marginal decrease in efficiency in single-tenant envs (where the tenant on the instance would match the one on `stripes` because, duh, there is only one tenant). 

Effectively, this new `isInstanceLoading` conditional makes the fallback irrelevant. We could prune the dead code if we wanted to.

PR #3019, opened 2026-05-13
UIIN-3642: ECS | Show child and parent of shared Instance when Instance from Member tenant

OleksandrHladchenko1:
## Purpose
* Show child and parent of shared Instance when Instance from Member tenant.

## Approach
* When `isInstanceShared` is true - pass central tenant id to the header of the request.

## Refs
[UIIN-3642](https://folio-org.atlassian.net/browse/UIIN-3642)

### DIFF
--- a/src/hooks/useInstancesQuery.js
+++ b/src/hooks/useInstancesQuery.js
@@ -32,7 +32,7 @@ const combineResults = (results) => {
 };
 
 // Fetches and returns multiple instances for given instance ids
-const useInstancesQuery = (ids = [], isInstanceShared) => {
+const useInstancesQuery = (ids = [], isInstanceShared = false) => {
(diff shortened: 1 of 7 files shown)


src/hooks/useInstancesQuery.js
```diff
@@ -32,8 +32,11 @@ const combineResults = (results) => {
 };
 
 // Fetches and returns multiple instances for given instance ids
-const useInstancesQuery = (ids = []) => {
-  const ky = useOkapiKy();
+const useInstancesQuery = (ids = [], isInstanceShared) => {
```
mariia-aloshyna: fix lint issue please

mariia-aloshyna (approved):

vashjs (approved):

PR #3023, opened 2026-05-28
[UIIN-3631] "Add Bound with and analytics" is disabled for new items

MikeTaylor:
This button is enabled when editing an existing item, but not when creating a new one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Source not displaying in Inventory version history for circ actions taken in different tenant. Fixes UIIN-3617.
 * Fetch items when holding accordion is open. Fixes UIIN-3658.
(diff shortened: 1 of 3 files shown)


zburke (changes requested):
If you can't add bound-withs until after the item is created, then the bound-with button on the create-item form should be omitted, not disabled. 

A form element should only be disabled when there is some action a user can take on the present page to enable it. e.g. a form's submit button is disabled while the form is pristine and then becomes enabled once the form is dirty. 

Sorry to knock your PR when it's the story that is incorrectly written. I know we had this convention documented in the old UX guidelines site but I'm struggling to find it in our storybook now. @JohnC-80 do know if/where this is written down?

BogdanDenis (approved):

zburke (approved):
Thanks for seeing this through while I put you all through the wringer on UX conventions 🤝

PR #3037, opened 2026-07-20
UIIN-3678 Add settings for number generator identifier

elsenhans:
[UIIN-3678](https://folio-org.atlassian.net/browse/UIIN-3678)

## Purpose

For enabling the new number generator for Instance identifier in Settings > Inventory > Instance, Holdings, Items (see https://folio-org.atlassian.net/browse/UIIN-3677 ) a drop-down for Identifier needs to be displayed.

 **Scenario 1** 
for enabling the number generator for Instance identifier a new section Instance needs to be displayed above the section Holdings.
Heading = **Instances**

 **Scenario 2**
in this section Instance a drop-down menu is displayed for enabling the number generator for Identifier 
Heading of drop-down = **Identifier**

**Scenario 3**
the drop-down for identifier contains:
- Off 
- On, field editable
- On, field not editable 

With the same functionality (e.g. enabling number generator for Item barcode (Inventory)) now for Instance identifier field (Number generator Name = Inventory: Instance identifier / Code = `inventory_instanceIdentifier`)

- Off (the identifier can be filled manually only) 
- On, field editable (the identifier can be filled using the generator and be edited, or filled manually)
- On, field not editable (the identifier can be filled using the generator only)

**Scenario 4**
no new capability/capability set is needed
existing capability/capability set
- application: app-inventory
- type: settings
- resource: UI-Inventory Settings Manage-Number-Generator-Options
- action: Manage

has to be assigned to select options in the drop-down menu for Identifier in addition to the other drop-downs (one capability/capability set for editing all drop-downs at Settings > Inventory > Instance, Holdings, Items)

## Approach
- Add Accordion for **Instance**
- Add Select field **Identifier** with options `dataOptionsAllEnabled` and default value `OFF`
- Add `const IDENTIFIER_SETTING = 'identifier';` and use it for `id` and `name` of the Field
- Add translation `"numberGenerator.identifier": "Identifier"`
- Add tests for Accordion and the input field

## Screenshots

image


src/settings/NumberGeneratorSettings/NumberGeneratorSettingsForm.js
```diff
@@ -139,7 +157,7 @@ const NumberGeneratorSettingsForm = ({
               </Row>
             </Accordion>
             <Accordion
-              id="acc02"
+              id="acc03"
```
zburke: Consider whether a more "semantic" identifier scheme would be better than this serial scheme that requires renaming every subsequent accordion when you push one onto the start of the list.

zburke (approved):

src/settings/NumberGeneratorSettings/NumberGeneratorSettingsForm.js
elsenhans: This option seems to be the preferred solution at Inventory. That's why we used it in first place.
I changed it to:
`id="accordion-instances"`, `id="accordion-holdings"` and `id="accordion-items"`

OleksandrHladchenko1 (approved):

PR #3038, opened 2026-07-20
UIIN-3650: Update Instance status term field when opening instance after updating

OleksandrHladchenko1:
## Purpose
* The "Instance status term" is displayed when opening the instance in Inventory app after it updating.

## Approach
* Update dependencies

## Refs
https://folio-org.atlassian.net/browse/UIIN-3650


zburke (approved):
Nice surgical fix, good. 

... But why are we bothering with memoization here in the first place? The performance gain from optimizing this tiny function cannot have been huge, and the tradeoff of the added complexity was a bug.

BogdanDenis (approved):

PR #3043, opened 2026-07-29
UIIN-3685: Update aria-label for action menu buttons based on context

zepheiryan:
## Purpose

Improve accessibility so users can differentiate between the purpose of each "Action" menu button despite the shared display text.

## Approach

- Update `SearchAndSave` to take an argument allowing properties to be set on the action menu button (waiting on this to merge in order to finish PR)
- Update `Pane`s or `DropdownButton`s to add the appropriate `aria-label`

## Refs

https://folio-org.atlassian.net/browse/UIIN-3685

### Related PRs

https://github.com/folio-org/stripes-smart-components/pull/1639

## Screenshots

Item details Actions button with updated aria-label:
Screenshot 2026-08-07 at 11 25 58

Results list Actions button with updated aria-label:
Screenshot 2026-08-07 at 11 25 33


src/components/InstancesList/InstancesList.js
```diff
@@ -1424,6 +1424,9 @@ class InstancesList extends React.Component {
           <SearchAndSort
             key={searchAndSortKey}
             actionMenu={this.getActionMenu}
+            actionMenuToggleProps={{
```
zepheiryan, line 1427: This prop is defined in https://github.com/folio-org/stripes-smart-components/pull/1639 and probably requires a stripes-smart-components release before this change can be incorporated into a release.

OleksandrHladchenko1 (approved):

